PR #1, opened 2017-04-06
set panter rails deploy

chrysalag:


config/deploy.rb
```diff
@@ -0,0 +1,5 @@
+# config valid only for current version of Capistrano
+lock "3.7.2"
```
kitsane: Those lines above can be deleted

kitsane:
After the change, you can rebase this branch onto development and merge it into development

PR #21, opened 2017-04-25
Feat/translation

chrysalag:
## Story
https://trello.com/c/S0RgDX57/26-as-a-user-i-want-the-application-to-be-translated-to-german

## Goals
- [ ] Story is tested
- [X] Acceptance criterias are met
- [ ] Code is reviewed
- [X] Travis is green
- [X] Story is on resolved in trello
- [X] Readme is updated if needed
- [ ] Styled with Bootstrap v3
- [ ] Mobile view is usable
- [ ] Seeds are available
- [X] German translations available


app/views/profiles/show.html.slim
```diff
@@ -1,35 +1,35 @@
...
+        td= _('Profession')
         td= @profile.profession
 
-    h2 Working days
+    h2= _('Working days')
     ul.list-inline
       - [:monday, :tuesday, :wednesday, :thursday, :friday].each do |day|
         - if @profile[day.to_s]
-          li #{day.to_s.titlecase}
+          li= t("weekdays.#{day.to_s}")
```
kitsane: Maybe `li= _(day.to_s)`?

app/views/users/_form.html.slim
```diff
@@ -13,4 +13,5 @@
 
   .row
     .col-xs-12
-      = f.button :submit, class: 'btn btn-default'
+    = f.button :submit, ( f.object.new_record? ? _('Create User') : _('Update User')),
```
kitsane: I really don't like the buttons this way. If there's no other way then this, i whould recomment to use standard rails translations for those cases.

Maybe it can be done through model attributes http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models

locale/app.pot
```diff
@@ -0,0 +1,253 @@
+# SOME DESCRIPTIVE TITLE.
```
kitsane: Why are there an app.pot and a app.edit.pot, can't there just be one? This will stay empty forever otherwise

locale/de/app.po
```diff
@@ -0,0 +1,251 @@
+# SOME DESCRIPTIVE TITLE.
```
kitsane: Shouldn't everything in the same app.po file? And why is there an app.po app.pot and app.edit.pot?

app/controllers/users_controller.rb
```diff
@@ -18,7 +18,8 @@ def create
     respond_to do |format|
       if @user.save
         @user.invite!
-        format.html { redirect_to users_path, notice: "Invitation sent to #{@user.email}" }
+        msg = format(_('Invitation sent to %{email}'), email: @user.email)
+        format.html { redirect_to users_path, notice: msg }
```
chrysalag: Rubocop complained about line length, that is the reason I did like this.

app/views/profiles/_form.html.slim
```diff
@@ -9,16 +9,17 @@
...
     .col-xs-12
-      h2 Arbeitstage
+      h2 =_('Working days')
       ul.list-inline
         - [:monday, :tuesday, :wednesday, :thursday, :friday].each do |day|
           li= f.input day, as: :boolean
   .row
     .col-xs-12
-      = f.button :submit, class: 'btn btn-default'
+      = f.button :submit, ( f.object.new_record? ? _('Create Profile') : _('Update Profile')),
```
chrysalag: Because the form is rendered in two views, each one has a different button and I needed the text of each to be translated.

app/views/profiles/show.html.slim
chrysalag: Parser doesn't find it that way.

PR #43, opened 2017-05-10
WIP: As a Superadmin, I want to create departments

casaper:
[Link to story in trello](#)

- [ ] Story under test
- [ ] Mobile works
- [ ] Seeds created
- [ ] Translated

Review by Team
- [ ] Reviewed by tnt
- [ ] Reviewed by cla

Review by Coach
- [x] Reviewed by sta

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170510150439) do
+ActiveRecord::Schema.define(version: 20170511135554) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -38,6 +38,8 @@ ActiveRecord::Schema.define(version: 20170510150439) do
     t.bigint "user_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.datetime "deleted_at"
+    t.index ["deleted_at"], name: "index_clients_on_deleted_at"
     t.index ["user_id"], name: "index_clients_on_user_id"
   end
 
@@ -47,7 +49,9 @@ ActiveRecord::Schema.define(version: 20170510150439) do
     t.string "level"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.datetime "deleted_at"
     t.index ["client_id"], name: "index_language_skills_on_client_id"
+    t.index ["deleted_at"], name: "index_language_skills_on_deleted_at"
   end
 
   create_table "profiles", force: :cascade do |t|
@@ -68,6 +72,8 @@ ActiveRecord::Schema.define(version: 20170510150439) do
     t.string "avatar_content_type"
     t.integer "avatar_file_size"
     t.datetime "avatar_updated_at"
+    t.datetime "deleted_at"
+    t.index ["deleted_at"], name: "index_profiles_on_deleted_at"
     t.index ["user_id"], name: "index_profiles_on_user_id"
   end
 
@@ -79,7 +85,9 @@ ActiveRecord::Schema.define(version: 20170510150439) do
     t.string "relation"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.datetime "deleted_at"
     t.index ["client_id"], name: "index_relatives_on_client_id"
+    t.index ["deleted_at"], name: "index_relatives_on_deleted_at"
   end
 
   create_table "schedules", force: :cascade do |t|
@@ -89,7 +97,9 @@ ActiveRecord::Schema.define(version: 20170510150439) do
     t.boolean "available", default: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.datetime "deleted_at"
     t.index ["client_id"], name: "index_schedules_on_client_id"
+    t.index ["deleted_at"], name: "index_schedules_on_deleted_at"
   end
 
   create_table "users", force: :cascade do |t|


config/locales/en.yml
```diff
@@ -222,3 +232,14 @@ en:
   voluntary_platform: Voluntaries Platform
   when: Time
   working_days: Working days
+  toggle_navigation: Toggle navigation menu
+  department_created: Department was successfully created.
+  department_updated: Department was successfully updated.
+  department_destroyed: Department was successfully deleted.
+  department_confirm_destroy: Are you sure you want to delete the departement?
```
chrysalag: department :)

.rubocop.yml
```diff
@@ -31,6 +31,7 @@ Style/AsciiComments:
 Metrics/LineLength:
   Max: 100
 Metrics/MethodLength:
+  Max: 20
```
kitsane: Please add it inline to the prams thing, because, globally i don't want methods that long

app/controllers/departments_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  private
+
+  def set_department
+    @department = Department.find(params[:id])
+  end
+
+  def department_params
+    email_attrs = [
```
kitsane: you could move this into separate method and which just return the array and then call the method

app/models/contact_point.rb
```diff
@@ -0,0 +1,3 @@
+class ContactPoint < ApplicationRecord
+  belongs_to :contact, optional: true
```
kitsane: add a validation, that body needs to be present and remove the `optional: true`

app/views/contacts/show.html.slim
```diff
@@ -0,0 +1,50 @@
...
+        dt= t_attr(:birthplace)
+        dd= @contact.birthplace
+
+      dt= t_attr(:organisation)
+      dd= @contact.organisation
+
+      dt= t_attr(:org_role)
+      dd= @contact.org_role
+
+- if kind == 'person'
```
kitsane: This fields shouldn't be in here, they are only on the volunteer, so please remove them

app/views/departments/_contact_fields.slim
```diff
@@ -0,0 +1,9 @@
+.nested-fields
+  .row
+    .col-xs-12.col-md-4= f.input :organisation
+    .col-xs-12.col-md-4= f.input :org_role
```
kitsane: drop org_role, no need for this

config/i18n-tasks.yml
```diff
@@ -0,0 +1,121 @@
+# i18n-tasks finds and manages missing and unused translations: https://github.com/glebm/i18n-tasks
```
kitsane: This file doesn't belongs_to this pull request, please delete

config/locales/de.yml
```diff
@@ -74,21 +103,20 @@ de:
...
-      - Mai
-      - Juni
-      - Juli
-      - August
-      - September
-      - Oktober
-      - November
-      - Dezember
+  month_names:
+    -
```
kitsane: Is there a reason for the empty line?

db/migrate/20170505141654_create_departments.rb
```diff
@@ -0,0 +1,9 @@
+class CreateDepartments < ActiveRecord::Migration[5.1]
+  def change
+    create_table :departments do |t|
+      t.string :name
```
kitsane: couldn't this be the contact company field?

test/factories/users.rb
```diff
@@ -9,7 +9,7 @@
     trait :with_clients do
       client_amt = 5
       clients do |client|
-        1.step(client_amt).to_a.map { |n| client.association(:client) }
```
kitsane: This file has nothing to do with your changes, why is it edited?

test/models/relative_test.rb
```diff
@@ -1,7 +1,7 @@
 require 'test_helper'
 
 class RelativeTest < ActiveSupport::TestCase
-  context "associations" do
+  context 'associations' do
```
kitsane: We don't use shoulda matchers for tests, so this test can be deleted

db/migrate/20170510080224_create_contacts.rb
```diff
@@ -0,0 +1,32 @@
...
+      t.string :last_name
+      t.string :organisation
+      t.string :kind
+      t.string :org_role
+      t.string :title         # function in organisation
+      t.string :url
+      t.string :birthplace
+      t.text :expertise
+      t.text :hobby
+      t.text :interest
```
kitsane: We should just focus on contact informations, so drop `:expertise, :hobby, :interest` and for what is `:org_role`and `:kind`?
`:title` i think is overkill, should also be deleted

app/views/departments/_form.html.slim
```diff
@@ -0,0 +1,21 @@
...
+  - if f.error_notification.present?
+    .row
+      .col-xs-12
+        = f.error_notification
+  .row
+    .col-xs-12
+      = f.input :name
+  .row
+    .col-xs-12
+      = f.association :user, as: :check_boxes, collection: User.department_assocable
```
nuritnt: I didn't understand, why we need here a checkbox with Users

app/controllers/departments_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def show
+    authorize @department
+  end
+
+  def new
+    @department = Department.new
+    @department.build_contact
+    @department.contact.contact_emails.build
+    @department.contact.contact_phones.build
```
kitsane: Here a build method in the model whould be good, something:
```ruby
def build_contact_with_points
  build_contact
  contact.contact_emails.build
  contact_phones.build
end
```

app/helpers/i18n_short_hand_helper.rb
```diff
@@ -0,0 +1,194 @@
+module I18nShortHandHelper
```
kitsane: This should be a pull request on @huerlisi gem.
Fork it, add it to your fork, make a pull request for simon and include the i18n_rails_helper gem via github from your fork

app/models/contact_email.rb
```diff
@@ -0,0 +1,5 @@
+class ContactEmail < ContactPoint
+  def self.label_collection
```
kitsane: `self.for_select`

app/models/contact_point.rb
casaper: doesnt work here. Everything breaks...

app/views/contacts/show.html.slim
casaper: here? This is in the contact controllers views and for later use... ;) so nope. denyed removal

db/migrate/20170510080224_create_contacts.rb
casaper: as our clients will use exactly these fields they're totally needed and will have to be in there. They reflect the naming used by the vCards standard....

app/helpers/i18n_short_hand_helper.rb
casaper: it is.. but untill that is thru, I want my views to be working

db/migrate/20170505141654_create_departments.rb
casaper: would be a kind of tricky thing. no? How do I validate presence in departement, but not in clients or profiles later on?

app/models/contact_email.rb
casaper: so why where the roles in user to be named role_collection and now there is a unspeaking for_select necessary? What if there would be more than one in the same model?

test/factories/users.rb
casaper: Because I've discovered the much nicer and more obvious way to create the mappable array with `Array.new(amt)` doing the other factory. So it makes sense to change it everywhere it is used to the more reasonable way.

app/views/contacts/show.html.slim
kitsane: Those fields are not part of your story, so drop them. @chrysalag  has them allready in her story for scaffolding volunteers. Also those fields are not part of contact informations. Later on when we have an address book, they wont have those fields too.

If we need them unified, we still can do an additional story. But that is not part of your story.
Oh and you don't decide what gets removed and what not, that is not your role.

app/models/contact_point.rb
kitsane: Hmm, ah because they get saved through the polymorphic. Sorry my mistake

app/models/contact_email.rb
kitsane: I don't remember the rails convention for the naming of this stuff, but yeah your right. Maybe you can find out. I think there is a 'for' in the name, but i don't remember

.rubocop.yml
```diff
@@ -30,9 +30,16 @@ Style/AsciiComments:
   Enabled: false
 Metrics/LineLength:
   Max: 100
+  Exclude:
+    - 'Gemfile'
+Metrics/ClassLength:
+  Exclude:
+    - 'test/*'
 Metrics/MethodLength:
+  Max: 20
```
kitsane: Drop this, as methods shouldn't be that long

app/controllers/contacts_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  def new
+    @contact = Contact.new
+  end
+
+  def edit; end
+
+  def create
+    @contact = Contact.new(contact_params)
+
+    respond_to do |format|
```
kitsane: Simplify to
```ruby
if @contact.save
  redirect_to @contact, notice: t('contact_created')
else
  render :new
end
```
The whole `respond_to` is not needed, when we only respond to html.

app/controllers/profiles_controller.rb
```diff
@@ -6,7 +6,7 @@ def show
   end
 
   def new
-    @profile = Profile.new(user_id: current_user.id)
```
kitsane: If you touch code outside of your story submit it as seperate merge request, because i want you to focus on your story

app/views/contacts/show.html.slim
```diff
@@ -0,0 +1,50 @@
+.row
+  .col-xs-12
+    h1= t_show
+
+.row
+  .col-xs-12
+    dl.dl-horizontal
```
kitsane: Why are you using dl, dt and dd tags? The other show views are solved with table

app/views/departments/show.html.slim
```diff
@@ -0,0 +1,49 @@
+.row
+  .col-xs-12
+    h1= @department.contact.org_role
+.row
+  .col-xs-12
+    dl.dl-slim
```
kitsane: Again why not a table?

db/migrate/20170510080224_create_contacts.rb
kitsane: We don't use vcards standard, otherwise we whould extend has_vcards but we decided to use our own, so we decide what fields are going to be in.

test/system/departments_test.rb
```diff
@@ -0,0 +1,158 @@
+require 'application_system_test_case'
+
+class DepartmentsTest < ApplicationSystemTestCase
+  def setup
+    @user = create :user, :with_profile, :with_clients,
+      :with_departments, role: 'superadmin'
+    @user_as_social_worker = create :user, :with_profile, :with_clients,
```
kitsane: `@social_worker` is enough, the user_as part doesn't add any value

test/system/departments_test.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    assert page.has_select? 'User'
+    within '.department_user' do
+      User.department_assocable.each do |u|
+        select u.email
+      end
+    end
+
+    assert page.has_field? 'Department name'
+    fill_in 'Department', with: 'bogus Department'
```
kitsane: The `assert page.has_field? 'Department name'` is not needed, as this line should fail otherwise

app/views/contacts/show.html.slim
casaper: > Acceptance Criteria:
> - Models: Contact, Address, ContactPoint
> .....

So If there was this scaffold in the story and we where talking about this beforhand, you're pissing me off pretty badly if you're comming allong now and tell me that I should not create a contact model and told chrisa to scaffold another contact model at the same time.
👎

app/models/contact_email.rb
casaper: well, the rails convention would be enum, but .... .... ....

app/views/contacts/show.html.slim
casaper: Table are for tabular data and not for key: value representations...

test/application_system_test_case.rb
```diff
@@ -1,7 +1,7 @@
 require 'test_helper'
 require 'capybara/poltergeist'
 
-DIMENSIONS = [1400, 1400].freeze
+DIMENSIONS = [1200, 3000].freeze
```
casaper: because test will fail with poltergeist but not with chrome... that simple. Fix Poltergeist, and I'll change them back

kitsane:
This pull request is way to good to be closed

app/views/contacts/show.html.slim
chrysalag: @casaper I am not scaffolding another contact model. Volunteers have for the time being address fields, email and phone number, but that is the same as it is now with clients.

But, I think @kitsane is talking about the expertise, hobby and interest fields. I have similar fields, but those aren't contact fields, are they?

PR #45, opened 2017-05-10
WIP: Features/deactivate users

nuritnt:
[Link to story in trello](#)

- [x] Story under test
- [x] Mobile works
- [x] Seeds created
- [x] Translated

Review by Team
- [x] Reviewed by cla
- [ ] Reviewed by vok

Review by Coach
- [x] Reviewed by sta


app/views/users/index.html.slim
```diff
@@ -9,6 +9,7 @@
         tr
           th= t_attr(:email)
           th= t_attr(:role)
+          th=
```
kitsane: remove `=`

app/views/users/show.html.slim
```diff
@@ -7,3 +7,7 @@
   .col-md-4
     strong= t_attr(:role)
   .col-md-8 = t("role.#{@user.role}")
+
+.row
+  .col-xs-12
+    p= link_to t('back'), users_path, class: 'btn btn-default'
```
kitsane: `p` is not needed

test/system/users_test.rb
```diff
@@ -64,4 +64,12 @@ class UsersTest < ApplicationSystemTestCase
     email = ActionMailer::Base.deliveries.last
     assert_equal 'superadmin@test.ch', email['to'].to_s
   end
+
+  test 'superadmins have no destroy link' do
```
kitsane: Call the test "superadmins can't destroy superadmins"

Gemfile
```diff
@@ -22,6 +22,7 @@ gem 'jbuilder'
 gem 'jquery-rails'
 gem 'panter-rails-deploy'
 gem 'paperclip'
+gem 'paranoia', '~> 2.2'
```
kitsane: Remove version number and rebundle again

kitsane:
Please add a model test, that ensures, that the record still exists after deletion

test/models/user_test.rb
```diff
@@ -74,4 +74,18 @@ def setup
   context 'associations' do
     should have_many(:clients)
   end
+
+  test 'record still exists after deletion' do
+    User.last.destroy
```
kitsane:
```ruby
user = create :user
user.destroy
assert_not user.find(user.id)
assert user.unscoped.find(user.id)
```

test/policies/user_policy_test.rb
```diff
@@ -8,6 +8,23 @@ def setup
     @user_as_department_manager = create :user, :with_profile, role: 'department_manager'
   end
 
+  test 'Destroy: superadmin can delete other user' do
+    assert_permit @user_as_superadmin, @user_as_social_worker, 'destroy?'
+    assert_permit @user_as_superadmin, @user_as_department_manager, 'destroy?'
+    assert_permit @user_as_other_superadmin, @user_as_social_worker, 'destroy?'
```
chrysalag: Maybe the check if an other superadmin can delete the entries another superadmin can is not needed. What is important, is that superadmin cannot delete another superadmin or herself. Exactly what you test in the next two tests.

test/policies/user_policy_test.rb
```diff
@@ -8,6 +8,23 @@ def setup
...
+    assert_permit @user_as_other_superadmin, @user_as_department_manager, 'destroy?'
+  end
+
+  test 'Destroy: superadmin cannot delete other superadmin' do
+    refute_permit @user_as_superadmin, @user_as_other_superadmin, 'destroy?'
+    refute_permit @user_as_other_superadmin, @user_as_superadmin, 'destroy?'
```
chrysalag: I think that line is also not needed. The two superadmins are created in the same way, so I think only the first check is needed.

test/policies/user_policy_test.rb
```diff
@@ -8,6 +8,23 @@ def setup
...
+  end
+
+  test 'Destroy: superadmin cannot delete itself' do
+    refute_permit @user_as_superadmin, @user_as_superadmin, 'destroy?'
+    refute_permit @user_as_other_superadmin, @user_as_other_superadmin, 'destroy?'
```
chrysalag: Same as above.

test/system/users_test.rb
```diff
@@ -4,7 +4,7 @@ class UsersTest < ApplicationSystemTestCase
   setup do
     @emails = ActionMailer::Base.deliveries
     @emails.clear
-    @user = create :user, email: 'superadmin@example.com'
+    @user = create :user, email: 'superadmin@example.com', role: 'superadmin'
```
chrysalag: The create :user, as is currently defined, creates by default a superadmin. So unless we need to set another role for the test, the role: is not needed.

test/system/users_test.rb
```diff
@@ -64,4 +64,12 @@ class UsersTest < ApplicationSystemTestCase
     email = ActionMailer::Base.deliveries.last
     assert_equal 'superadmin@test.ch', email['to'].to_s
   end
+
+  test "superadmins can't destroy superadmin" do
```
chrysalag: I think that this test needs to be changed.

PR #54, opened 2017-05-19
As a Department Manager and Superadmin, I want to manage Departments

casaper:
As a Department Manager and Superadmin, I want to manage Departments

- Department Managers can't change the assigned Department Manager
- As Department Manager there is a link in the Navbar to his department
- Department Managers can update the department's fields

[Link to story in trello](https://trello.com/c/dtIXQFBa/13-as-a-department-manager-and-superadmin-i-want-to-manage-departments-sp-5)

- [x] Story under test
- [x] Mobile works
-  ~~Seeds created~~
-  ~~Translated~~

Review by Team
- [x] Reviewed by tnt
- [x] Reviewed by cla
- [ ] Reviewed by vok

Review by Coach
- [ ] Reviewed by sta

### DIFF
--- a/app/controllers/departments_controller.rb
+++ b/app/controllers/departments_controller.rb
@@ -5,9 +5,7 @@ class DepartmentsController < ApplicationController
     @departments = policy_scope(Department)
   end
 
-  def show
-    authorize @department
-  end
+  def show; end
 
   def new
     @department = Department.new
@@ -15,9 +13,7 @@ class DepartmentsController < ApplicationController
     authorize @department
   end
 
-  def edit
-    authorize @department
-  end
+  def edit; end
 
   def create
     @department = Department.new
@@ -31,7 +27,6 @@ class DepartmentsController < ApplicationController
   end
 
   def update
-    authorize @department
     if @department.update_attributes(permitted_attributes(@department))
       redirect_to @department, notice: t('department_updated')
     else
@@ -40,7 +35,6 @@ class DepartmentsController < ApplicationController
   end
 
   def destroy
-    authorize @department
     @department.destroy
     redirect_to departments_url, notice: t('department_destroyed')
   end
@@ -49,9 +43,6 @@ class DepartmentsController < ApplicationController
 
   def set_department
     @department = Department.find(params[:id])
-  end
-
-  def department_params
-    params.require(:department).permit(policy(@department).permitted_attributes)
+    authorize @department
   end
 end
--- a/app/controllers/departments_controller.rb
+++ b/app/controllers/departments_controller.rb
@@ -20,7 +20,8 @@ class DepartmentsController < ApplicationController
   end
 
   def create
-    @department = Department.new(department_params)
+    @department = Department.new
+    @department.update_attributes(permitted_attributes(@department))
     authorize @department
     if @department.save
       redirect_to @department, notice: t('department_created')
@@ -31,7 +32,7 @@ class DepartmentsController < ApplicationController
 
   def update
     authorize @department
-    if @department.update(department_params)
+    if @department.update_attributes(permitted_attributes(@department))
       redirect_to @department, notice: t('department_updated')
     else
       render :edit
@@ -50,24 +51,7 @@ class DepartmentsController < ApplicationController
     @department = Department.find(params[:id])
   end
 
-  def phone_attrs
-    [:id, :body, :label, :_destroy, :type, :contacts_id]
-  end
-
-  def email_attrs
-    [:id, :body, :label, :_destroy, :type, :contacts_id]
-  end
-
-  def contact_attrs
-    [
-      :id, :name, :_destroy, :contactable_id, :contactable_type, :street,
-      :extended, :city, :postal_code,
-      contact_emails_attributes: email_attrs,
-      contact_phones_attributes: phone_attrs
-    ]
-  end
-
   def department_params
-    params.require(:department).permit(user_ids: [], contact_attributes: contact_attrs)
+    params.require(:department).permit(policy(@department).permitted_attributes)
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -32,6 +32,10 @@ class User < ApplicationRecord
     superadmin? || department_manager?
   end
 
+  def department?
+    department_manager? && department.first.present?
+  end
+
   def self.create_user_and_send_password_reset(email:, role:)
     new_user = User.new(
       email: email, password: Devise.friendly_token, role: role
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -32,7 +32,7 @@ class User < ApplicationRecord
     superadmin? || department_manager?
   end
 
-  def department?
+  def with_department?
     department.first.present?
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -33,7 +33,7 @@ class User < ApplicationRecord
   end
 
   def department?
-    department_manager? && department.first.present?
+    department.first.present?
   end
 
   def self.create_user_and_send_password_reset(email:, role:)
--- a/app/policies/department_policy.rb
+++ b/app/policies/department_policy.rb
@@ -31,7 +31,7 @@ class DepartmentPolicy < ApplicationPolicy
   end
 
   def show?
-    superadmin_or_users_department?
+    superadmin_or_the_departments_manager?
   end
 
   def new?
@@ -39,7 +39,7 @@ class DepartmentPolicy < ApplicationPolicy
   end
 
   def edit?
-    superadmin_or_users_department?
+    superadmin_or_the_departments_manager?
   end
 
   def create?
@@ -55,7 +55,7 @@ class DepartmentPolicy < ApplicationPolicy
   end
 
   def update?
-    superadmin_or_users_department?
+    superadmin_or_the_departments_manager?
   end
 
   def destroy?
@@ -64,7 +64,7 @@ class DepartmentPolicy < ApplicationPolicy
 
   private
 
-  def superadmin_or_users_department?
+  def superadmin_or_the_departments_manager?
     @user.superadmin? || @department.user.include?(@user)
   end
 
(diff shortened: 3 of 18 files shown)


app/views/application/_navigation.html.slim
```diff
@@ -19,6 +19,8 @@ nav.navbar.navbar-top
           li= link_to t('volunteers'), volunteers_path
           - if policy(Department).index?
             li= link_to t('departments'), departments_path
+          - if current_user.department?
```
chrysalag: Shouldn't it be something like if policy(Department).show?

test/models/department_test.rb
```diff
@@ -2,13 +2,12 @@
...
+    @department = create :department
   end
 
-  test 'department is saveable with user relation' do
-    refute @department.valid?
-    @department.contact = build :contact
+  test 'department is valid' do
     assert @department.valid?
-    assert @department.save!
+    @department.contact.name = ''
```
chrysalag: shouldn't this and the following line be in a separate test, like test 'department without required fields is invalid'?

test/system/departments_test.rb
```diff
@@ -118,6 +123,56 @@ def setup
...
+  test "Department Managers can update their department's fields" do
+    login_as @department_manager
+    visit edit_department_path(@department_manager.department.first.id)
+    refute page.has_select? 'User'
+    fill_in 'Name', with: 'Name changed'
+    fill_in 'Street', with: 'Street changed'
+    fill_in 'Extended address', with: 'Extended address changed'
+    fill_in 'Zip', with: 'Zip changed'
+    fill_in 'City', with: 'City changed'
+    within '#phones' do
```
chrysalag: I think there should be a fill_in here.

app/views/application/_navigation.html.slim
casaper: This checks if the user has a department and if the user is a department_manager. But yea, maybe this should be in the policy. But I'm just not sure how... I'll find out.

test/models/department_test.rb
casaper: your right, but not. It should go, but it shouldn't be in this model test at all. Plus I've allready tested for this in contact_test before. And thats where it should be, as it is a contact, not a department concern. Thx.

test/system/departments_test.rb
casaper: no, but this
```ruby
within '#phones' do
```
shouldn't be there at all, as the `#phones` field is filled in at line 155 ;)

config/locales/en.yml
```diff
@@ -76,6 +76,7 @@ en:
       - :month
       - :year
   department_created: Department was successfully created.
+  department_updated: Department was successfully updated.
   departments: Departments
```
chrysalag: should there also be a department_destroyed? I get en.department_destroyed translation missing at the develop

app/policies/department_policy.rb
```diff
@@ -40,6 +63,14 @@ def create?
     @user.superadmin?
   end
 
+  def associate_user?
```
kitsane: As you just check for `@user.superadmin?` in the method, why not just use `@user.superadmin?`

app/controllers/departments_controller.rb
```diff
@@ -50,24 +51,7 @@ def set_department
...
-      :id, :name, :_destroy, :contactable_id, :contactable_type, :street,
-      :extended, :city, :postal_code,
-      contact_emails_attributes: email_attrs,
-      contact_phones_attributes: phone_attrs
-    ]
-  end
-
   def department_params
-    params.require(:department).permit(user_ids: [], contact_attributes: contact_attrs)
+    params.require(:department).permit(policy(@department).permitted_attributes)
```
kitsane: Not my prefered way as, but @chrysalag @nuritnt should decide on this

app/policies/department_policy.rb
```diff
@@ -20,6 +20,29 @@ def initialize(user, department)
...
+
+  def contact_attrs
+    [
+      :id, :name, :_destroy, :contactable_id, :contactable_type, :street,
+      :extended, :city, :postal_code,
+      contact_emails_attributes: email_attrs,
+      contact_phones_attributes: phone_attrs
+    ]
+  end
+
```
nuritnt: @kitsane can we do this week a coaching session with the topic related to these lines 23-44:
1. Action Controller Parameters
2. technical question: why is it better to move these lines in the policy instead in a concern? or what is the difference here?

test/models/department_test.rb
```diff
@@ -2,13 +2,10 @@
 
 class DepartmentTest < ActiveSupport::TestCase
   def setup
-    @department = Department.new
+    @department = create :department
```
kitsane: A `build :department` is enough

app/models/user.rb
```diff
@@ -32,6 +32,10 @@ def superadmin_or_department_manager?
     superadmin? || department_manager?
   end
 
+  def department?
```
kitsane: There is the method `manager_with_department?`, but i'm thinking, that this method should be named that way and the other `department?`, your opinion?

app/policies/department_policy.rb
```diff
@@ -20,6 +20,29 @@ def initialize(user, department)
     @department = department
   end
 
+  def permitted_attributes
+    return [user_ids: [], contact_attributes: contact_attrs] if @user.superadmin?
+    return [contact_attributes: contact_attrs] if @user.department_manager?
+    []
```
nuritnt: i didnt understand why return is used here and the empty brackets.

app/models/user.rb
```diff
@@ -32,6 +32,10 @@ def superadmin_or_department_manager?
...
+    department.first.present?
+  end
```
nuritnt: I didn't understand why we use here .first here?

app/controllers/departments_controller.rb
nuritnt: tbh I had difficulties to follow or understand this line..

app/models/user.rb
```diff
@@ -32,6 +32,10 @@ def superadmin_or_department_manager?
     superadmin? || department_manager?
   end
 
+  def department?
+    department.first.present?
```
casaper: because the department field gives back an array, so present? would not work on that.

app/policies/department_policy.rb
casaper: return terminates the method. So if the condition is fullfilled the array is returned and the stuff afterwards is not going to be executed.

app/policies/department_policy.rb
casaper: In the policy I can assign different permited parameters based on the logged in user role. So in this case the department manager is not allowed to set user_ids to the department, but the superadmin is.

app/policies/department_policy.rb
casaper: can you clarify this? I don't understand what you mean.

app/models/user.rb
casaper: I did name it longer first, and the linter suggested me to shorten its name. But yeah, I'm willing to change the name if people prefer it to be changed.

app/controllers/departments_controller.rb
casaper: @nuritnt where where you're difficulties? 
`policy(@department).permitted_attributes` returns the array with allowed parameters. The method is [here](https://github.com/panter/aoz-003/blob/d4924f04aff1927fd0b272ce0bd5726041e7ba51/app/policies/department_policy.rb#L23)

@kitsane how would you evaluate user's permission to set attributes, if not like this?

casaper: and not evaluating such things is just very bad practice.
If I don't evaluate this, a department_manager can just add a select manually to his department form within his browsers inspect:
```html
<select name="department[user_ids][]" id="department_user_ids">
  <option value="666"></option>
</select>
```
Submission will just be accepted and commited to the database. Maybe rails will evaluate that the id 666 is not existing. But if I hit an existing id, it will be in the database.
This is just the simplest method, there would be much tougher stuff that could be done. This could be accomplished by anyone that knows what a Browsers inspector is and knows how to insert a select into his form there.

You can well argue that this is an edge case, and you want to trust the users and so on. 
But this is just **extremely bad practice** and sooner or later it will flash back with a breach that will be at least very embarrassing, if not harmful.

casaper: PS: To demonstrate this I've just added a comment to https://aoz-003-staging.panter.biz/clients/7 as a socialworker. This action whould simply destroy all the text that is in there, if any, and place the text of my choice in there.

casaper: And as a social_worker role user I shouldn't be able to do so.

app/views/application/_navigation.html.slim
casaper: What do you think about the way I did it now?

app/models/user.rb
casaper: this is what the linter thinks about `has_department?`
> [Style/PredicateName] Style/PredicateName: Rename `has_department?` to `department?`.

But `with_department?` doesn't seem to upset it.
`manager_with_department?` is in my view too long and `@user.manager_with_department?` is doublemeaning, because as far as I can see a `@user` should only be assignable to departments if `@user.role == 'department_manager'`. But `with_department?` looks quite all right to me. What do you guys think?

app/policies/department_policy.rb
```diff
@@ -20,31 +20,66 @@ def initialize(user, department)
...
   end
 
   def destroy?
     @user.superadmin?
   end
+
+  private
+
+  def superadmin_or_users_department?
+    @user.superadmin? || @department.user.include?(@user)
```
kitsane: `@user.superadmin? || manager_with_department?` and the name of the method `superadmin_or_department_manager?`, maybe?

app/controllers/departments_controller.rb
kitsane: It's ok to include this mechanism, please explain to the time the how.

app/policies/department_policy.rb
```diff
@@ -20,31 +20,66 @@ def initialize(user, department)
     @department = department
   end
 
+  def permitted_attributes
```
kitsane: Please add a test to this method

kitsane:
Retrigger travis, to get green, looks good.

casaper:
ready to rebase?

PR #59, opened 2017-05-26
Process volunteers

chrysalag:
[Link to story in trello](https://trello.com/c/A91i0pkG/15-as-a-superadmin-i-want-to-process-volunteers-sp-3)

- [ ] Story under test
- [ ] Mobile works
- [ ] Seeds created
- [ ] Translated

Review by Team
- [x] Reviewed by tnt
- [ ] Reviewed by cla
- [ ] Reviewed by vok

Review by Coach
- [x] Reviewed by sta

### DIFF
--- a/app/assets/javascripts/rejection.js
+++ b/app/assets/javascripts/rejection.js
@@ -1,0 +1,17 @@
+$(function() {
+  show_rejection();
+
+  $('#volunteer_state').change(function(){
+    show_rejection();
+  });
+});
+
+function show_rejection() {
+  if($('#volunteer_state').val() == 'rejected') {
+    $('.volunteer_rejection_type').show();
+    $('.volunteer_rejection_text').show();
+  } else {
+    $('.volunteer_rejection_type').hide();
+    $('.volunteer_rejection_text').hide();
+  }
+}
--- a/app/assets/javascripts/rejection.js
+++ b/app/assets/javascripts/rejection.js
@@ -1,8 +1,10 @@
 $(function() {
-  show_rejection();
-
-  $('#volunteer_state').change(function(){
+  $( document ).on('turbolinks:render, turbolinks:load', function() {
     show_rejection();
+
+    $('#volunteer_state').on('change', function(){
+      show_rejection();
+    });
   });
 });
 
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -1,7 +1,6 @@
 class VolunteersController < ApplicationController
   include NestedAttributes
   before_action :set_volunteer, only: [:show, :edit, :update, :destroy]
-  before_action :set_state, only: [:update]
 
   def index
     @volunteers = Volunteer.all
@@ -27,10 +26,10 @@ class VolunteersController < ApplicationController
 
   def update
     if @volunteer.update(volunteer_params)
-      if @state == 'interested' && @volunteer.state == 'accepted'
+      if @volunteer.state == 'accepted'
         new_user = User.new(email: @volunteer.email,
           password: Devise.friendly_token, role: 'volunteer')
-        new_user.save!
+        new_user.save
         new_user.invite!
         redirect_to volunteers_path, notice: t('invite_sent', email: new_user.email)
       else
@@ -50,10 +49,6 @@ class VolunteersController < ApplicationController
 
   def set_volunteer
     @volunteer = Volunteer.find(params[:id])
-  end
-
-  def set_state
-    @state = @volunteer.state
   end
 
   def volunteer_params
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -1,7 +1,7 @@
 class VolunteersController < ApplicationController
   include NestedAttributes
   before_action :set_volunteer, only: [:show, :edit, :update, :destroy]
-  before_action :set_state, only: [:set_volunteer, :update]
+  before_action :set_state, only: [:update]
 
   def index
     @volunteers = Volunteer.all
@@ -27,7 +27,7 @@ class VolunteersController < ApplicationController
 
   def update
     if @volunteer.update(volunteer_params)
-      if @state == 'interested' && @volunteer.state == 'active'
+      if @state == 'interested' && @volunteer.state == 'accepted'
         new_user = User.new(email: @volunteer.email,
           password: Devise.friendly_token, role: 'volunteer')
         new_user.save!
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -1,6 +1,7 @@
 class VolunteersController < ApplicationController
   include NestedAttributes
   before_action :set_volunteer, only: [:show, :edit, :update, :destroy]
+  before_action :set_state, only: [:update]
 
   def index
     @volunteers = Volunteer.all
@@ -26,7 +27,7 @@ class VolunteersController < ApplicationController
 
   def update
     if @volunteer.update(volunteer_params)
-      if @volunteer.state == 'accepted'
+      if @state == Volunteer::INTERESTED && @volunteer.state == Volunteer::ACCEPTED
         new_user = User.new(email: @volunteer.email,
           password: Devise.friendly_token, role: 'volunteer')
         new_user.save
@@ -46,6 +47,10 @@ class VolunteersController < ApplicationController
   end
 
   private
+
+  def set_state
+    @state = @volunteer.state
+  end
 
   def set_volunteer
     @volunteer = Volunteer.find(params[:id])
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -5,19 +5,26 @@ class VolunteersController < ApplicationController
 
   def index
     @volunteers = Volunteer.all
+    authorize Volunteer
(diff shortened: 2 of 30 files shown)


config/locales/de.yml
```diff
@@ -403,6 +404,7 @@ de:
           department_manager: Standort Verantwortliche/r
           social_worker: SozialarbeiterIn
           superadmin: Superadmin
+          #volunteer: Freiwillige
```
nuritnt: why is this commented out?

chrysalag: Because I forgot to delete it.

app/models/user.rb
```diff
@@ -4,13 +4,16 @@ class User < ApplicationRecord
...
+  has_one :volunteer, dependent: :destroy
   has_and_belongs_to_many :department
 
   # Roles definition
   SUPERADMIN = 'superadmin'.freeze
   SOCIAL_WORKER = 'social_worker'.freeze
   DEPARTMENT_MANAGER = 'department_manager'.freeze
-  ROLES = [SUPERADMIN, SOCIAL_WORKER, DEPARTMENT_MANAGER].freeze
+  VOLUNTEER = 'volunteer'.freeze
+  ROLES = [SUPERADMIN, SOCIAL_WORKER, DEPARTMENT_MANAGER, VOLUNTEER].freeze
```
kitsane: `ROLES = ROLES_FORM + [VOLUNTEER]`

app/controllers/volunteers_controller.rb
```diff
@@ -26,7 +27,15 @@ def create
 
   def update
     if @volunteer.update(volunteer_params)
-      redirect_to @volunteer, notice: t('volunteer_updated')
+      if @state == 'interested/registered' && @volunteer.state == 'active'
+        new_user = User.new(email: @volunteer.email,
+          password: Devise.friendly_token, role: 'volunteer')
+        new_user.save!
+        new_user.invite!
```
kitsane: Why is the password stuff needed? In the users_controller, they just get invited without setting password.
Also i don't see here that the volunteer gets associated to the user_account, something like:
```ruby
@volunteer.build_user(email: @volunteer.email)
```
And all this should be done before `if @volunteer.update(volunteer_params)`.

app/controllers/volunteers_controller.rb
```diff
@@ -26,7 +27,15 @@ def create
 
   def update
     if @volunteer.update(volunteer_params)
-      redirect_to @volunteer, notice: t('volunteer_updated')
+      if @state == 'interested/registered' && @volunteer.state == 'active'
```
kitsane: You double check here the same. So just use `@volunteer.state`
```diff
+  def set_state
+    @state = @volunteer.state
+  end
```

app/controllers/volunteers_controller.rb
```diff
@@ -1,6 +1,7 @@
 class VolunteersController < ApplicationController
   include NestedAttributes
   before_action :set_volunteer, only: [:show, :edit, :update, :destroy]
+  before_action :set_state, only: [:set_volunteer, :update]
```
kitsane: Why is this action called before `:set_volunteer`

app/models/user.rb
```diff
@@ -4,13 +4,16 @@ class User < ApplicationRecord
...
+  ROLES_FORM = [SUPERADMIN, SOCIAL_WORKER, DEPARTMENT_MANAGER].freeze
```
kitsane: `ROLES_FOR_COLLECTION`maybe?

app/views/volunteers/_form.html.slim
```diff
@@ -58,9 +58,10 @@
     = f.input :skills, label: t('your_skills')
     = f.input :interests, label: t('your_interests')
 
-.row
-  .col-xs-12
-    = f.input :state, as: :hidden
+- if !current_user.nil? && current_user.superadmin?
```
kitsane: `if current_user.present? && current_user.superadmin?` or use the policy to check

app/views/volunteers/edit.html.slim
```diff
@@ -2,7 +2,12 @@
   .col-xs-12
     h1= t_title(:edit)
 
-== render 'form'
+.row
+  .col-xs-12
+  = simple_form_for(@volunteer) do |f|
```
kitsane: This `form` is outside of the `.col-xs-12`

app/controllers/volunteers_controller.rb
chrysalag: state is the old state, volunteer.state is the state after the update.

app/controllers/volunteers_controller.rb
chrysalag: In the create action of the users controller, a user with password is created before user.save and user.invite

app/assets/javascripts/rejection.js
```diff
@@ -0,0 +1,17 @@
+$(function() {
```
kitsane: Why not using bootstrap collapse?

chrysalag: Didn't know it existed. I am still looking to find out how to add data toggle & target to the corresponding collection value. Why is the use of javascript problematic?

kitsane: No it is not, just thougt bootstrap has already a nice solution, this is why i mention it, could be a working solution

app/controllers/volunteers_controller.rb
```diff
@@ -25,20 +33,34 @@ def create
   end
 
   def update
+    authorize @volunteer
     if @volunteer.update(volunteer_params)
-      redirect_to @volunteer, notice: t('volunteer_updated')
+      if @state == Volunteer::INTERESTED && @volunteer.state == Volunteer::ACCEPTED
```
kitsane: Just use `@volunteer.state`

app/views/volunteers/_form.html.slim
```diff
@@ -58,9 +58,14 @@
     = f.input :skills, label: t('your_skills')
     = f.input :interests, label: t('your_interests')
 
-.row
-  .col-xs-12
-    = f.input :state, as: :hidden
+- if current_user.present? && policy(Volunteer).destroy?
```
casaper: `current_user.present?` can be avoided by doing this
```ruby
def destroy?
  user && user.superadmin?
end
```
in the `VolunteerPolicy`

so you only need
```ruby
if policy(Volunteer).destroy?
```

app/views/volunteers/edit.html.slim
```diff
@@ -2,7 +2,12 @@
   .col-xs-12
     h1= t_title(:edit)
 
-== render 'form'
+.row
+  .col-xs-12
+    = simple_form_for(@volunteer) do |f|
```
casaper: Dont nest this form in `.row .col-xs-12` because it is not needed -> div soup

casaper: PS: the form partial has .row .col-xs-12 again, so all that form rows and columns will be nested within this col-xs-12

chrysalag: If I remove the nesting then there is no space between the update and show buttons.

app/models/user.rb
```diff
@@ -4,13 +4,16 @@ class User < ApplicationRecord
...
+  has_one :volunteer, dependent: :destroy
   has_and_belongs_to_many :department
 
   # Roles definition
   SUPERADMIN = 'superadmin'.freeze
   SOCIAL_WORKER = 'social_worker'.freeze
   DEPARTMENT_MANAGER = 'department_manager'.freeze
-  ROLES = [SUPERADMIN, SOCIAL_WORKER, DEPARTMENT_MANAGER].freeze
+  VOLUNTEER = 'volunteer'.freeze
+  ROLES_FOR_COLLECTION = [SUPERADMIN, SOCIAL_WORKER, DEPARTMENT_MANAGER].freeze
```
chrysalag: You're right. Since volunteers will have an account, therefore access to the system, I would suggest ROLES_FOR_USER_CREATE or ROLES_FOR_CREATE.

app/controllers/volunteers_controller.rb
chrysalag: `@volunteer.state` contains the state after the update. `@state` contains the state before the update. We need to check, if the state has changed from interested to accepted. How can I check that by only using `@volunteer.state'?

app/views/volunteers/edit.html.slim
casaper: I see what you mean. But I think there is better ways to solve this. 

Just as a possible example amongst many, change `_content.scss` to: 
```scss
.row,
.simple_form {
  margin-bottom: $padding-large-horizontal;
}
```

casaper: Im stating this, because I realized that all of us (including myself) where extensiveley nesting things. At first this works just fine, but as soon these nestings are getting messed up (they will), it tends to get hairy...

app/assets/javascripts/rejection.js
kitsane: I really miss the resolve feature from gitlab. This is resolved

kitsane:
I whould like to give this branch over to @nuritnt to finish and merge

app/controllers/volunteers_controller.rb
chrysalag: @kitsane So, does this need to change or not?

PR #82, opened 2017-06-21
Generic form navigation links (as buttons) helper

casaper:
Review by Team
- [x] Reviewed by tnt
- [ ] Reviewed by cla

Review by Coach
- [ ] Reviewed by sta

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -28,4 +28,17 @@ module ApplicationHelper
     html_options[:partial] = "#{association}/fields"
     link_to_add_association(name, f, association, html_options)
   end
+
+  def form_navigation_btn(action)
+    text = action == :back ? t('back') : t_title(action)
+    action = :index if action == :back
+    target = { controller: controller_name, action: action }
+    button_type = action == :new ? 'success' : 'default'
+    target[:id] = params[:id] unless action == :index
+    content_tag :div, class: 'row' do
+      content_tag :div, class: 'col-xs-12' do
+        link_to text, target, class: "btn btn-#{button_type}"
+      end
+    end
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -23,6 +23,10 @@ module ApplicationHelper
     single_col_xs f.error_notification if f.error_notification.present?
   end
 
+  def button_link(text, target, type = 'default')
+    link_to text, target, class: "btn btn-#{type}"
+  end
+
   def link_to_add_polymorphic_association(*args)
     name, f, association, html_options = *args
     html_options[:partial] = "#{association}/fields"
@@ -35,10 +39,6 @@ module ApplicationHelper
     target = { controller: controller_name, action: action }
     button_type = action == :new ? 'success' : 'default'
     target[:id] = params[:id] unless action == :index
-    content_tag :div, class: 'row' do
-      content_tag :div, class: 'col-xs-12' do
-        link_to text, target, class: "btn btn-#{button_type}"
-      end
-    end
+    single_col_xs button_link(text, target, button_type)
   end
 end
--- a/app/views/clients/edit.html.slim
+++ b/app/views/clients/edit.html.slim
@@ -4,6 +4,5 @@
 
 == render 'form'
 
-.row
-  .col-xs-12
-    = link_to t('back'), clients_path, class: 'btn btn-default'
+= form_navigation_btn :show
+= form_navigation_btn :back
--- a/app/views/clients/edit.html.slim
+++ b/app/views/clients/edit.html.slim
@@ -4,5 +4,4 @@
 
 == render 'form'
 
-= form_navigation_btn :show
 = form_navigation_btn :back
--- a/app/views/departments/edit.html.slim
+++ b/app/views/departments/edit.html.slim
@@ -4,6 +4,5 @@
 
 == render 'form'
 
-.row
-  .col-xs-12
-    = link_to t(:back), departments_path, class: 'btn btn-default'
+= form_navigation_btn :show
+= form_navigation_btn :back
--- a/app/views/departments/edit.html.slim
+++ b/app/views/departments/edit.html.slim
@@ -4,5 +4,4 @@
 
 == render 'form'
 
-= form_navigation_btn :show
 = form_navigation_btn :back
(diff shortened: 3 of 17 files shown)


app/views/application/_button_back.html.slim
```diff
@@ -0,0 +1,3 @@
+.row
+  .col-xs-12
+    = link_to t('back'), { controller: controller_name, action: :index }, { class: 'btn btn-default' }
```
kitsane: Check if {} are needed

app/views/clients/_form.html.slim
```diff
@@ -63,6 +63,4 @@
 
     = render 'schedules/form', f: f, participant: @client
 
-  .row
-    .col-xs-12
-      = f.button :submit, class: 'btn btn-success'
+  = render 'button_submit', f: f
```
kitsane: Check if is configurable through simple_form

app/views/application/_button_edit.html.slim
```diff
@@ -0,0 +1,3 @@
+.row
+  .col-xs-12
+    = link_to t_action(:edit), { controller: controller_name, action: :edit, id: params[:id] }, { class: 'btn btn-default' }
```
nuritnt: can't follow here. 

old version:
`link_to t_action(:edit), edit_client_path(@client), class: 'btn btn-default'`

with this new way - how does it know to go on which path? 
and the `@client` is here then directed with the` id: params[:id]` ?

need explanations here for better understanding the refactoring

app/views/application/_button_back.html.slim
casaper: they're needed

app/views/application/_button_edit.html.slim
casaper: Because this partial wants to be generic, we do not have edit_client_path( nor `@client`. Or we have it, but it could be `@client`, `@volunteer`, `@user`, .... depending on which show-view it is being rendered on.

This way we can just generate the correct path from the parameters and the responsible controller. It means that this partial would even work if we'd make a custom resource path, that isn't named edit_xyz_path.

casaper:
Excluded submit button from this PR and did all within a generic helper. Please review again, because it's a different thing now...

app/views/volunteers/edit.html.slim
```diff
@@ -4,6 +4,5 @@ h1= t_title(:edit)
   = render 'volunteers/form', f: f, volunteer: @volunteer
   = f.button :submit, class: 'btn btn-success'
 
-.row
-  .col-xs-12
-    = link_to t('back'), volunteers_path, class: 'btn btn-default'
+= form_navigation_btn :show
```
nuritnt: same here, show button here probably not needed, right?

app/views/profiles/show.html.slim
```diff
@@ -34,10 +34,8 @@
               - else
                 i.glyphicon.glyphicon-remove.text-danger
                 span.sr-only Doesn't work on #{day.to_s}s
-.row
-  .col-xs-12
-    = link_to t('edit_profile'), edit_profile_path(@profile), class: 'btn btn-default'
 
+= form_navigation_btn :edit
 .row
```
nuritnt: is there a possibility to use this refactoring for this button?

app/views/departments/edit.html.slim
```diff
@@ -4,6 +4,5 @@
 
 == render 'form'
 
-.row
-  .col-xs-12
-    = link_to t(:back), departments_path, class: 'btn btn-default'
\ No newline at end of file
+= form_navigation_btn :show
```
nuritnt: same here, show button here probably not needed, right?

app/views/clients/edit.html.slim
```diff
@@ -4,6 +4,5 @@
 
 == render 'form'
 
-.row
-  .col-xs-12
-    = link_to t('back'), clients_path, class: 'btn btn-default'
+= form_navigation_btn :show
```
nuritnt: you replaced a `back` button with `show` & `back`.
I think the show button is here not needed, right?
and I didn't understand where this `show` button in the application helper is defined as the others are

app/helpers/application_helper.rb
```diff
@@ -16,4 +16,17 @@ def link_to_add_polymorphic_association(*args)
     html_options[:partial] = "#{association}/fields"
     link_to_add_association(name, f, association, html_options)
   end
+
+  def form_navigation_btn(action)
+    text = action == :back ? t('back') : t_title(action)
+    action = :index if action == :back
+    target = { controller: controller_name, action: action }
+    button_type = action == :new ? 'success' : 'default'
+    target[:id] = params[:id] unless action == :index
```
nuritnt: I couldn't totally understand the single line conditionals from line 21 - 25,
can we have a look together on this

app/views/departments/edit.html.slim
casaper: It is there due to the fact that it hase been build upon a state you havent removed that button. And now it is here because of rebasing

app/views/profiles/show.html.slim
casaper: not without refactoring the helper. it uses non standard text and a target path that is not related to its views controller.

app/helpers/application_helper.rb
casaper: These are [Ternary if or Conditional ?:](http://rubylearning.com/satishtalim/simple_constructs.html), not single line conditionals. 
```ruby
def form_navigation_btn(action)
  text = action == :back ? t('back') : t_title(action)
  # equals
  if (action == :back)
    text = t('back')
  else
    text = t_title(action)
  end

  action = :index if action == :back
  # equals
  if(action == :back)
    action = :index
  end

  target = { controller: controller_name, action: action }

  button_type = action == :new ? 'success' : 'default'
  # equals
  if(action == :new )
    button_type = 'success'
  else
    button_type = 'default'
  end

  target[:id] = params[:id] unless action == :index
  single_col_xs button_link(text, target, button_type)
end
```

app/views/departments/edit.html.slim
nuritnt: you probably meant the opposite - when I check out at https://github.com/panter/aoz-003/commit/02712307e311c5c8913a01cb683647d6661b816e - and I remember deleting it for the related files.
but anyway - do we need/does it make sense to have the show button in this view or what do you think?

app/views/profiles/show.html.slim
nuritnt: ah I see

app/helpers/application_helper.rb
nuritnt: ah thanks! will have a look on that! 🤓

app/views/departments/edit.html.slim
nuritnt: nevermind, saw your changes after I replied ;)

PR #87, opened 2017-06-30
Example for pdf generation

kitsane:
Only as example for setup and how to generate a pdf, not for merge


test/models/assignment_test.rb
```diff
@@ -0,0 +1,7 @@
+require 'test_helper'
+
+class AssignmentTest < ActiveSupport::TestCase
+  # test "the truth" do
+  #   assert true
+  # end
+end
```
alesrosina: You can remove this file, since it is empty.

app/controllers/clients_controller.rb
```diff
@@ -44,6 +44,11 @@ def destroy
     redirect_to clients_url, make_notice
   end
 
+  def without_assignments
```
alesrosina: Maybe rename to `need_accompanying`, so you would have nicer routes file (see also comment for `routes.rb`)

app/controllers/assignments_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    client.state = Client::REGISTERED
+    client.save
+    if Volunteer.without_clients.include?(volunteer)
+      volunteer.state = Volunteer::ACCEPTED
+      volunteer.save
+    end
+    redirect_to assignments_url, make_notice
+  end
+
+  def find_volunteer
```
alesrosina: Maybe move to `ClientController`, so you would have nicer routes file (see also comment for routes.rb).

config/routes.rb
```diff
@@ -1,4 +1,8 @@
 Rails.application.routes.draw do
+  get '/clients/need_accompanying', to: 'clients#without_assignments'
+  get '/clients/find_volunteer', to: 'assignments#find_volunteer'
```
chrysalag: I have them like this because it was in the requirements to have such urls.

PR #90, opened 2017-07-01
remove duplicate method single_col_xs

casaper:
minor bug fix

Review by Team
- [ ] Reviewed by tnt
- [ ] Reviewed by cla
- [ ] Reviewed by vok

Review by Coach
- [ ] Reviewed by sta

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -11,14 +11,6 @@ module ApplicationHelper
     end
   end
 
-  def single_col_xs(inside, cols = 12)
-    content_tag :div, class: 'row' do
-      content_tag :div, class: "col-xs-#{cols}" do
-        inside
-      end
-    end
-  end
-
   def simple_error_notice(f)
     single_col_xs f.error_notification if f.error_notification.present?
   end


app/helpers/application_helper.rb
```diff
@@ -11,14 +11,6 @@ def week(with_weekend = false)
...
-  def single_col_xs(inside, cols = 12)
-    content_tag :div, class: 'row' do
-      content_tag :div, class: "col-xs-#{cols}" do
-        inside
-      end
-    end
-  end
-
   def simple_error_notice(f)
     single_col_xs f.error_notification if f.error_notification.present?
```
chrysalag: Since single_col_xs function is removed, shouldn't also the calls to it get removed? (There is also a call to it at the application helper.)

casaper: it is duplicated. same method twice in the helpers module

chrysalag: Ah! I am sorry!

casaper: no worries.

PR #95, opened 2017-07-06
Journal über Klienten oder Freiwillige

casaper:
[Link to story in trello](https://trello.com/c/L9K0jum4/3-epic-journal-%C3%BCber-klienten-oder-freiwillige)

- [x] Story under test
- [x] Mobile works
- [x] Seeds created
- [x] Translated

Review by Team
- [ ] Reviewed

### DIFF
--- a/app/controllers/journals_controller.rb
+++ b/app/controllers/journals_controller.rb
@@ -40,6 +40,8 @@ class JournalsController < ApplicationController
     redirect_to @journaled, make_notice
   end
 
+  private
+
   def journal_relations
     {
       journalable_type: @journaled.class.model_name.name,
@@ -47,8 +49,6 @@ class JournalsController < ApplicationController
       user_id: current_user.id
     }
   end
-
-  private
 
   def set_journal
     @journal = Journal.find(params[:id])
@@ -62,7 +62,7 @@ class JournalsController < ApplicationController
 
   def journal_params
     params.require(:journal).permit(
-      :subject, :user_id, :body, :client_id, :volunteer_id, :client_id
+      :subject, :user_id, :body, :client_id, :volunteer_id
     )
   end
 end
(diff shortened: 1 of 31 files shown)


app/controllers/journals_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+      render :edit
+    end
+  end
+
+  def destroy
+    @journal.destroy
+    redirect_to @journaled, make_notice
+  end
+
+  def journal_relations
```
alesrosina: I guess this one could be moved to private.

app/models/journal.rb
```diff
@@ -0,0 +1,3 @@
+class Journal < ApplicationRecord
+  belongs_to :user
```
alesrosina: Is this correct? You have defined `has_many :journals` in `client`, not in `user`.

casaper: The user writes a journal entry about the client. So the user is the author, and it is about the client.

app/controllers/journals_controller.rb
casaper: yea, probably your right.

app/models/journal.rb
alesrosina: OK, makes sense. I will also add `belongs_to :journalable, polymorphic: true` in `Journal` and `has_many :journals` to `User`, so both relations are defined on a model.

app/controllers/journals_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    {
+      journalable_type: @journaled.class.model_name.name,
+      journalable_id: @journaled.id,
+      user_id: current_user.id
+    }
+  end
+
+  def set_journal
+    @journal = Journal.find(params[:id].to_i)
```
alesrosina: To integer transformation is not necessary, Rails is smart enough - so `Journal.find("1")` is the same as `Journal.find(1)`. Try it out in console, it will work. Will remove it before merging it.

PR #110, opened 2017-07-12
matching volunteer side

chrysalag:
# Pull Request

## Done?

- [ ] Acceptance criterias are met
- [ ] Code quality checks are green
- [ ] Readme updated if needed
- [ ] Story under test
- [ ] Mobile works
- [ ] Seeds created
- [ ] Translated
- [ ] Code is reviewed

- [ ] Deployed to staging after mergeing

### DIFF
--- a/app/views/assignments/_age_request_select.html.slim
+++ b/app/views/assignments/_age_request_select.html.slim
@@ -1,4 +1,4 @@
-- age_request_filter = request_filter(:age_request_cont)
+- age_request_filter = request_filter(:age_request_cont, 'all')
 .btn-group
   button.btn.btn-default.dropdown-toggle type='button' data-toggle='dropdown' aria-label="#{t_attr(:age_request)}" aria-haspopup='true' aria-expanded='false' class="btn-md"
     = "#{t_attr(:age_request, Client)} : "
--- a/app/views/assignments/_age_request_select.html.slim
+++ b/app/views/assignments/_age_request_select.html.slim
@@ -1,12 +1,11 @@
-- age_request_filter = params[:q] && params[:q][:age_request_cont].present? ? params[:q][:age_request_cont] : 'age_no_matter'
+- age_request_filter = params[:q] && params[:q][:age_request_cont].present? ? params[:q][:age_request_cont] : 'all'
 .btn-group
   button.btn.btn-default.dropdown-toggle type='button' data-toggle='dropdown' aria-label="#{t_attr(:age_request)}" aria-haspopup='true' aria-expanded='false' class="btn-md"
     = "#{t_attr(:age_request, Client)} : "
     = t(age_request_filter, scope: [:simple_form, :options, :client, :age_request]) + ' '
     span.caret aria-hidden='true'
   ul.dropdown-menu
+    li
+      = link_to t('all'), params.permit!.to_h.deep_merge(q: {age_request_cont: ''})
     - Client.age_request_collection.each do |s|
-      - if s.to_s == 'age_no_matter'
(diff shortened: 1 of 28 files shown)


config/locales/de.yml
```diff
@@ -341,6 +342,7 @@ de:
         one: 'Konnte %{resource} nicht speichern: ein Fehler.'
         other: 'Konnte %{resource} nicht speichern: %{count} Fehler.'
   entry_year: Einreisejahr
+  find_client: KlientIn finden
```
nuritnt: I think we decided to write the genders like this: Klient/in

toupeira: Why not simply "Klient"? 👿 

Food for thought: http://www.rhetorik.ch/Feministisch/Feministisch.html

nuritnt: we just took over the way aoz uses it, that we don't have to discuss personal/political topics 😄

toupeira: No worries, I'm okay with that ;-)

casaper: Klient/in is a matter of taste. But anyhow: It is a requirement from AOZ, so not our choice....

app/views/assignments/_age_request_select.html.slim
```diff
@@ -0,0 +1,11 @@
+- age_request_filter = params[:q] && params[:q][:age_request_cont].present? ? params[:q][:age_request_cont] : 'all'
```
toupeira: Complicated logic like this should be moved into a helper, also allows easier testing.

app/views/assignments/_gender_request_select.html.slim
```diff
@@ -0,0 +1,11 @@
+- gender_request_filter = params[:q] && params[:q][:gender_request_cont].present? ? params[:q][:gender_request_cont] : 'all'
```
toupeira: Same here and in `_language_skills_language_select.html.slim` below, you can make a generic helper which takes the query parameter as argument

app/views/assignments/_gender_request_select.html.slim
```diff
@@ -0,0 +1,11 @@
...
+.btn-group
+  button.btn.btn-default.dropdown-toggle type='button' data-toggle='dropdown' aria-label="#{t_attr(:gender_request)}" aria-haspopup='true' aria-expanded='false' class="btn-md"
+    = "#{t_attr(:gender_request, Client)} : "
+    = t(gender_request_filter, scope: [:simple_form, :options, :client, :gender_request]) + ' '
+    span.caret aria-hidden='true'
+  ul.dropdown-menu
+    li
+      = link_to t('all'), params.permit!.to_h.deep_merge(q: {gender_request_cont: ''})
```
toupeira: This params-merging looks like another candidate for a helper :)

app/views/assignments/find_client.html.slim
```diff
@@ -0,0 +1,91 @@
...
+.table-responsive
+  table.table.table-no-border-top
+    tbody
+      = render 'contacts/show', contact: @volunteer.contact
+      tr
+        td= t_attr(:date_of_birth)
+        td= l(@volunteer.date_of_birth) if @volunteer.date_of_birth
+      tr
+        td= t_attr(:gender)
+        td= t("gender.#{@volunteer.gender}") if @volunteer.gender.present?
```
toupeira: You can just do `gender?` instead of `gender.present?` (works for all model attributes, but not associations)

app/views/clients/_client.html.slim
```diff
@@ -16,12 +16,17 @@ tr
   td = client.goals
   td = client.competent_authority
   td = client.created_at.strftime("%d-%m-%Y")
+  - if params[:action] == 'find_client'
```
toupeira: Rails has a helper `action_name` which is slightly more robust.

PR #112, opened 2017-07-12
add attributes to assignments

chrysalag:
# Pull Request

## Done?

- [x] Acceptance criterias are met
- [x] Code quality checks are green
- ~[ ] Readme updated if needed~
- [x] Story under test
- [x] Mobile works
- ~[ ] Seeds created~
- [x] Translated
- [x] Code is reviewed

- [ ] Deployed to staging after merging

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1,6 +1,7 @@
 class Assignment < ApplicationRecord
   belongs_to :client
   belongs_to :volunteer
+  belongs_to :user
 
   def self.state_collection
     [:suggested, :active]
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1,4 +1,8 @@
 class Assignment < ApplicationRecord
   belongs_to :client
   belongs_to :volunteer
+
+  def self.state_collection
+    [:suggested, :active]
+  end
 end
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1,7 +1,7 @@
 class Assignment < ApplicationRecord
   belongs_to :client
   belongs_to :volunteer
-  belongs_to :user
+  belongs_to :creator, class_name: 'User', foreign_key: 'creator_id'
 
   def self.state_collection
     [:suggested, :active]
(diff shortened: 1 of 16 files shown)


config/locales/de.yml
```diff
@@ -5,6 +5,8 @@ de:
       assignment:
         agreement: Zustimmung
         client: KlientIn
+        created_by: Verantwortlich
```
nuritnt: Verantwortliche/r

app/models/assignment.rb
```diff
@@ -1,4 +1,9 @@
 class Assignment < ApplicationRecord
   belongs_to :client
   belongs_to :volunteer
+  belongs_to :user
```
nuritnt: for my understanding, why did we add here the user? for accessing the user_id right? is it better to access it directly via the user instead the user_id from the client/volunteer?

app/views/assignments/_form.html.slim
```diff
@@ -1,5 +1,6 @@
 = simple_form_for(@assignment) do |f|
   = simple_error_notice f
+  = f.input :user_id, as: :hidden
```
nuritnt: also for my understanding related to my question above, when we add here the hidden field, isn't the id passed with current "client" ?

toupeira: Not sure, but this field is ignored because the user is overriden in the controller:

```ruby
@assignment.user = current_user
```

It usually makes sense to do that, otherwise the user can theoretically change the ID in the hidden field and create a record for another user.

app/models/assignment.rb
chrysalag: The user_id of the assignment and the user_id of the client/volunteer might not be the same. user_id of the assignment is the user who created the assignment. Client user is the user who registered the client and Volunteer user is, if I remember correctly, the user who the volunteer becomes when she gets an account. So there is high chance they are not the same person.

app/views/assignments/_form.html.slim
chrysalag: Yes, you are right. The hidden field in not needed. I will remove it.

app/models/assignment.rb
toupeira: One option to make this clearer is to call the association `creator` instead, i.e. something like `belongs_to :creator, class_name: User` (= column name `creator_id`)

PR #100, opened 2017-07-08
Models that require Contact or Schedule to be built on them, take care of this themselves

casaper:
Just soing something that has always hurt me, when I looked at the thing. SCNR.

PR #115, opened 2017-07-14
WIP: Fixes/user profile relation

nuritnt:
# Pull Request

## Done?

- [ ] Acceptance criterias are met
- [ ] Code quality checks are green
- [ ] Readme updated if needed
- [ ] Story under test
- [ ] Mobile works
- [ ] Seeds created
- [ ] Translated
- [ ] Code is reviewed

- [ ] Deployed to staging after mergeing

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -12,8 +12,6 @@ class UsersController < ApplicationController
 
   def new
     @user = User.new
-    @user.build_profile
-    @user.profile.build_contact
     authorize @user
   end
 
@@ -21,11 +19,6 @@ class UsersController < ApplicationController
 
   def create
     @user = User.new user_params.merge(password: Devise.friendly_token)
-    # this needs to be done to ensure department handling dont fail
-    @user.build_profile
-    @user.profile.build_contact(contact_params)
-    @user.profile.contact.primary_email = @user.email
-
     if @user.save
       @user.invite!
       redirect_to users_path, notice: t('invite_sent', email: @user.email)
@@ -58,10 +51,10 @@ class UsersController < ApplicationController
   end
 
   def user_params
-    params.require(:user).permit(:email, :password, :role)
+    params.require(:user).permit(:email, :password, :role,
+      profile_attributes: [
+        contact_attributes: contact_attributes[:contact_attributes]
+      ])
   end
 
-  def contact_params
-    params.require(:contact).permit(contact_attributes[:contact_attributes])
-  end
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,4 +1,5 @@
 class UsersController < ApplicationController
+  include ContactAttributes
   before_action :set_user, only: [:show, :edit, :update, :destroy]
   after_action :verify_authorized
 
@@ -11,6 +12,8 @@ class UsersController < ApplicationController
 
   def new
     @user = User.new
+    @user.build_profile
+    @user.profile.build_contact
     authorize @user
   end
 
@@ -18,6 +21,11 @@ class UsersController < ApplicationController
 
   def create
     @user = User.new user_params.merge(password: Devise.friendly_token)
+    # this needs to be done to ensure department handling dont fail
+    @user.build_profile
+    @user.profile.build_contact(contact_params)
+    @user.profile.contact.primary_email = @user.email
+
     if @user.save
       @user.invite!
       redirect_to users_path, notice: t('invite_sent', email: @user.email)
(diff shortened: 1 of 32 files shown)


app/controllers/users_controller.rb
```diff
@@ -11,13 +12,20 @@ def show; end
 
   def new
     @user = User.new
+    @user.build_profile
+    @user.profile.build_contact
```
toupeira: The traditional Rails way for this is [nested attributes](https://rubyplus.com/articles/3681-Complex-Forms-in-Rails-5), but it has its downsides as well. Not saying you should change it in this PR, just something to read up on :)

toupeira: You can also do something similar as @casaper did in #100 to move this logic out of the controller.

test/factories/users.rb
```diff
@@ -20,5 +20,12 @@
         Array.new(1).map { |_n| create(:department) }
       end
     end
+
+    # after(:create) do |user, profile|
+    #   user.profile
+    #   user.reload
+    #   # u.profile = create(:profile, user: u)
+    #   # u.profile.contact = create(:contact, contactable: u)
+    # end
```
toupeira: You can try `after(:build)`, IIRC that would get called before the record is saved, as opposed to `after(:create)` which gets called after saving (which I assume causes the validation errors).

Or you can use the `after_initialize` approach from #100 then it should happen automatically.

casaper:
This thing has just loads and loads of trouble inside. It breaks basicly any tests everywhere, and if I try to fix on one side, i get 10 new failing tests at the other end. The construct with User.profile.contact and Volunteer.contact is just pure mess in my view.

casaper:
# Questions:

- Whats this PR's Story in Trello?
- What should it actually fix?
- For now I see only a fix that brings actually a feature (first and last_name fielsd in user creation form). So is it really a fix?

toupeira:
@nuritnt, @casaper ok I tried to fix the failing tests, hopefully I didn't break anything. I think I need someone to sit me down and explain the data model, I'm a bit confused to say the least :)

I'll leave some comments in my commit.

toupeira:
Also, the `BuildContactRelation` introduced in #100 is a bit problematic because an empty contact is always invalid, this should probably be refactored to use nested attributes as well.

casaper:
@toupeira I reckon the profile thingie should be dropped and User should have `has_one :contact` and `has_many :schedules` just like the volunteers and the clients. But the messy thing is that the volunteers are going thru that metamorphosis and turn into users. so they have two contact relations. All in all, I was getting a little frustrated with that thing today, to be honest. It was the perfect round up to basically any test we have in place, and pulling the string on one end just .... ah, I just loved the thing 👊 👊👊👊👊 😄

test/test_helper.rb
```diff
@@ -10,12 +10,15 @@ class ActiveSupport::TestCase
   DatabaseCleaner.strategy = :transaction
 
   def before_setup
+    # FIXME: make sure all users are deleted, sometimes records stick around
+    # when tests are aborted
+    User.with_deleted.map(&:really_destroy!)
```
casaper: I had trouble with database cleaner before. And it wasn't because of killing the tests. Misteriously multiplying records within `def setup`. E.g. 9 Users after creating one and such.
Thats why I did this "Holzhammermethode" thingie that was here in place before.

I didn't realize it was because of paranoia, but now it makes quite allot of sense. Thanks for the enlightenment.

app/models/user.rb
```diff
@@ -11,6 +10,8 @@ class User < ApplicationRecord
   has_one :volunteer, dependent: :destroy
   has_and_belongs_to_many :department
 
+  before_validation :assign_primary_email, if: :profile
```
casaper: Sorry for messing with that. I might have just accidently removed that when tests where blasting all the versions of red on me...

nuritnt:
@casaper I saw your questions comment, and don't know if you got it already answered or figured it out by yourself. 

> - What should it actually fix?
> - For now I see only a fix that brings actually a feature (first and last_name fielsd in user creation form). So is it really a fix?

The problem was, that when a user is created, there is no profile(&contact) created. And if you are prompted to enter a password on first login, you are not prompted to create your profile, so it will be empty. In some cases, as for example in department creation, we use all users to create one - which can be problematic, if one is empty - and it crashes. 

This is problem is a edge case, but nevertheless caused some problems, and was a tricky one to find out (at least for me).
In overall, as we have a lot of models and "tricky" relations, it is getting complex - and (if we would have time at some point) consider our data modelling and maybe refactor even some, as @toupeira already mentioned some concerns. This was also a feedback from @alesrosina I got, when he was helping me with figuring out this story.

🙏Tausenddank for your help guys.🚀 I hope you didn't get too much frustrated 😇 (like me on friday 😝 )

PR #122, opened 2017-07-19
Volunteer has many assignments

chrysalag:
# [Story in Trello](https://trello.com/c/FmcIBtN6/32-ein-freiwilliger-kann-mehrere-begleitungen-gleichzeitig-machen-3)
# [Story in Trello](https://trello.com/c/QZbHQ68y/31-bei-freiwilligen-die-begleitungen-anzeigen-2)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?


app/views/volunteers/show.html.slim
```diff
@@ -2,6 +2,11 @@
   .col-xs-12
     h1= @volunteer.contact.full_name
 
+- if Volunteer.seeking_clients.include?(@volunteer)
```
toupeira: Doing DB queries from a view should usually be avoided, and it's also not necessary here since you already have the volunteer's `state`. It's not a big deal, but you could e.g. add a `Volunteer::SEEKING_STATES` constant, use it in `Volunteer.seeking_clients`, add an instance method `Volunteer#seeking_clients?`, and then just do `@volunteer.seeking_clients?` here.

toupeira: BTW the `state` attributes in `Volunteer` and `Client` could be simplified by using [AR enums](http://www.justinweiss.com/articles/creating-easy-readable-attributes-with-activerecord-enums/), but that would also involve switching from strings to integers. Maybe something for a Kata session some day :)

PR #124, opened 2017-07-19
Fix PDF generation with superadmin that has no profile

casaper:
…t any profile

# [Story in Trello](https://trello.com/c/GHZeQZhk/49-check-after-deploy-exception-on-staging-when-creating-pdf)

### Done for approval?

* [ ] ~Readme updated if needed~
* [x] Story under test
* [ ] ~Mobile works~
* [ ] ~Seeds created~
* [ ] ~Translated~
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/views/assignments/show.html.slim
+++ b/app/views/assignments/show.html.slim
@@ -148,37 +148,38 @@
 
 .row
   .col-xs-11.col-xs-offset-1
-    .row
-      .col-xs-6
-        .row
-          .col-xs-5
-            = t('assignment_pdf.last_name')
-          .col-xs-7.col-input
-            = @assignment.creator.profile.contact.last_name
-        .row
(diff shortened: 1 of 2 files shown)


casaper:
@chrysalag 
But a user not having a profile at all, isn't it a edge case?
And If there'd be a profile with contact in it that has blank first_name and lastname etc. fields, this part of the template will still be run?

chrysalag:
@casaper It is indeed an edge case. But if it can happen, what I would suggest is that the fields for completion appear in all cases and they are filled out in case they have a value.

casaper:
@chrysalag I would rather suggest to fix that it can happen, so that this will only be a worst case catch

casaper:
@toupeira agreee. I'm glad to make one...

PR #126, opened 2017-07-19
Features/superuser can edit other users data

casaper:
# [Story in Trello](https://trello.com/c/JCRzRcI5/41-feedback-superuser-administration)

### Done for approval?

* [ ] Story under test
* [x] Mobile works
* [ ] Seeds created
* [ ] Translated
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -30,7 +30,7 @@ class UsersController < ApplicationController
 
   # only used to update the current user
   def update
-    if @user.update(user_params)
+    if handle_update(user_params)
       bypass_sign_in @user
       redirect_to @user, notice: t('profile_updated')
     else
@@ -45,6 +45,14 @@ class UsersController < ApplicationController
 
   private
 
+  def handle_update(user_params)
+    if user_params[:password].blank?
+      @user.update_without_password(user_params.slice(:email, :role))
+    else
+      @user.update(user_params)
+    end
+  end
+
   def set_user
     @user = User.find(params[:id])
     authorize @user
@@ -56,5 +64,4 @@ class UsersController < ApplicationController
         contact_attributes: contact_attributes[:contact_attributes]
       ])
   end
-
 end
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -31,7 +31,7 @@ class UsersController < ApplicationController
   # only used to update the current user
   def update
     if handle_update(user_params)
-      bypass_sign_in @user
+      bypass_sign_in @user if @user == current_user
       redirect_to @user, notice: t('profile_updated')
     else
       render :edit
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -47,7 +47,7 @@ class UsersController < ApplicationController
 
   def handle_update(user_params)
     if user_params[:password].blank?
-      @user.update_without_password(user_params.slice(:email, :role))
+      @user.update_without_password(user_params)
     else
       @user.update(user_params)
     end
--- a/app/models/contact.rb
+++ b/app/models/contact.rb
@@ -29,7 +29,7 @@ class Contact < ApplicationRecord
   end
 
   def validate_first_name?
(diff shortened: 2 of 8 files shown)


app/models/contact.rb
```diff
@@ -28,6 +28,14 @@ def needs_primary_email?
     ['Client', 'Volunteer'].include? contactable_type
   end
 
+  def validate_first_name?
+    contactable_type != 'Department' && contactable_type != 'Profile'
```
toupeira: Sorry looks like I got it wrong in #115 ;-)

toupeira: You can simplify it to `!department? && !profile?` though.

app/policies/user_policy.rb
```diff
@@ -33,4 +33,8 @@ def update?
   def destroy?
     @current_user.superadmin? && !@user.superadmin?
   end
+
+  def reset_password?
+    @current_user.superadmin? && @current_user != @user
```
toupeira: Not sure if this matches the requirement:

> Als Superuser möchte ich Passwort von anderen Superusern resetten (oder wenn einfacher überschreiben)

Shouldn't it be `@current_user.superadmin? || @current_user == @user`, i.e. superadmins can reset all passwords, other users only their own?

And yes this definitely needs a test :)

app/views/users/_form.html.slim
```diff
@@ -3,8 +3,9 @@
   .row
     .col-xs-12
       = f.input :email
-      - if current_user == user
+      - if current_user == user || policy(User).reset_password?
```
toupeira: If my assumption above is correct you can just do `if policy(User).reset_password?` here

app/models/contact.rb
casaper: I did it exactly like that at first, but it just behaved strange on me - didn't happen to work.
But no problem, I can try again.

app/policies/user_policy.rb
casaper: Nope. The user himself can do this within the login form with devises own password forgotten feature.

app/views/users/_form.html.slim
casaper: right! got me here 😉

app/models/contact.rb
casaper: > Sorry looks like I got it wrong in #115 ;-)

How. What should be validated, and what not? I thought we should un-restrict the thing for user.profile, in order to avoid exceptions thrown because of that?

casaper:
@toupeira 

> What about email addresses?

What is about them?

app/policies/user_policy.rb
```diff
@@ -33,4 +33,8 @@ def update?
   def destroy?
     @current_user.superadmin? && !@user.superadmin?
   end
+
+  def reset_password?
+    @current_user.superadmin? || @current_user == @user
+  end
```
toupeira: Just realized this is the same as `update?` above, is there a reason we need this?

toupeira: Also see https://github.com/panter/aoz-003/issues/132

casaper: yea, maybe in this case your right, because the update? would still be explicit. I'll change it. 

But in other cases `policy(Xy).destroy?` was used in views for things that where absolutely not related to `destroy?` abilities. 
That is, i reckon, not very explicit.

app/controllers/users_controller.rb
```diff
@@ -45,6 +45,14 @@ def destroy
 
   private
 
+  def handle_update(user_params)
+    if user_params[:password].blank?
+      @user.update_without_password(user_params.slice(:email, :role))
```
toupeira: `update_without_password` [already deletes the password attributes](https://github.com/plataformatec/devise/blob/71fc5b351a8e8473733b0ca29d60c06e43a0efe1/lib/devise/models/database_authenticatable.rb#L98-L100), so you can just pass all `user_params`

casaper: ah, didn't realize...
thx.

PR #134, opened 2017-07-26
Import clients from current AOZ MS-Access DB

casaper:
# [Story in Trello](https://trello.com/c/WY3ii7i6/59-import-clients-5)

* Country mapping will be solved in another story


lib/access_import/acc_utils.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def email(h_email)
+    return "unknown_email_#{Time.now.to_i}@example.com" if h_email.nil?
+    h_email.sub(/^\#mailto:/, '').sub(/\#$/, '')
+  end
+
+  def map_gender(value)
+    return nil unless value
+    return 'female' if value == 'W'
```
chrysalag: I think we don't have at all now female/male, we only have mrs/mr.

lib/access_import/access_import.rb
```diff
@@ -0,0 +1,34 @@
...
+    @personen_rollen.all_clients.each do |key, ac_client|
+      next if Client.where('access_import @> ?', {
+        id_personen_rolle: ac_client[:pk_PersonenRolle].to_i
+      }.to_json).any?
+      client_attrs = client_transformer.prepare_attributes(ac_client)
+      client = Client.new(client_attrs)
+      client.user = User.first
+      puts "Importing personen_rolle #{key} and Client.id: #{client.id}" if client.save!
+    end
+    puts "Imported #{Client.count - client_count_before} new clients from MS Access Datbase."
```
chrysalag: Database

lib/access_import/sprachen.rb
```diff
@@ -0,0 +1,34 @@
...
+  def sanitize_record(rec)
+    rec = parse_int_fields(rec, [:pk_Sprache])
+    {
+      lang: language(rec[:t_Sprache]),
+      sprache: rec[:t_Sprache]
+    }
+  end
+
+  def language(sprache)
+    sprache = 'Tigrinja' if sprache == 'Tigrinia'
```
chrysalag: I think it is Tigrin**y**a

lib/access_import/haupt_person.rb
```diff
@@ -0,0 +1,63 @@
...
+    end.to_h
+    @records = all
+  end
+
+  def all
+    internal_all if @records.blank?
+    @records
+  end
+
+  # reltions
```
chrysalag: relations

lib/access_import/acc_utils.rb
```diff
@@ -0,0 +1,54 @@
...
+    return 'male' if value == 'M'
```
nuritnt: Maybe I didn't see it so far or this will be changed later but we don't have any `gender` left except in the volunteer application as `gender request` everything was changed to mr/mrs with the salutation.

lib/access_import/begleitete.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def find(id)
+    internal_find(id)
+    @record
+  end
+
+  def sanitize_record(rec)
+    rec = parse_int_fields(rec, [:fk_FamilienRolle, :fk_PersonenRolle, :pk_Begleitete])
+    rec = parse_datetime_fields(rec, [:d_MutDatum])
+    rec[:gender] = map_gender(rec[:t_Geschlecht])
```
nuritnt: again, no gender left, only salutation

lib/access_import/client_transform.rb
```diff
@@ -0,0 +1,127 @@
...
+        first_name: relative[:t_Vorname],
+        last_name: relative[:t_Name], relation: relative[:relation],
+        birth_year: relative[:z_Jahrgang] && Date.parse("#{relative[:z_Jahrgang]}-01-01")
+      }]
+    end.to_h
+  end
+
+  def birth_year(geburtsdatum, jahrgang)
+    return geburtsdatum if geburtsdatum
+    Date.parse(jahrgang + '-06-01') if jahrgang
```
nuritnt: what does this `-06-01` do?

lib/access_import/haupt_person.rb
```diff
@@ -0,0 +1,63 @@
...
+    self
+  end
+
+  def internal_find(id)
+    if @records
+      @records[id.to_s]
+    else
+      person = @acdb[:tbl_Hauptpersonen].select do |hp|
+        hp[:pk_Hauptperson] == id.to_s
+      end.first
```
nuritnt: general question: what does dis `end.first` in do?

lib/access_import/haupt_person.rb
nuritnt: ? type? does this needs to be done or hä?

lib/access_import/sprachen.rb
casaper: Nope. I have researched. And thats where our gem matches.

lib/access_import/client_transform.rb
casaper: it just sets the date in the middle of the year. Because where only using the year, but have a full date field.

lib/access_import/haupt_person.rb
casaper: select returns an array, even if there is only one match. So with first I free it from beeing nested as the only element inside an array.

lib/access_import/sprachen.rb
casaper: Its german

casaper: well, maybe your right @chrysalag , but I still don't care. If that bothers you, you have to report an issue to the gem that matches on Tigrinja ;)

chrysalag: I am sorry, I had found a wikipedia article in German that had it like that and I thought it was like that. Didn't know that the gem has it otherwise.

lib/access_import/acc_utils.rb
casaper: A: I wrote that while we still had that gender field
B: it is quite possible that this will be needed to map other stuff in the future.

lib/access_import/sprachen.rb
casaper: I had to do this because the value that came from the access db was missspelled anyway. Thats why this is there at all.

lib/access_import/client_transform.rb
```diff
@@ -0,0 +1,127 @@
...
+      personen_rolle: down_hkeys(fields[2]),
+      familien_rolle: down_hkeys(fields[3].except(:d_MutDatum, :t_Mutation)),
+      land: fields[4]
+    }
+  end
+
+  def contact_attributes(haupt_person, plz)
+    mobile_number(haupt_person).merge(
+      first_name: haupt_person[:t_Vorname] || 'unbekannt',
+      last_name: haupt_person[:t_Nachname] || 'unbekannt',
```
toupeira: Not sure how empty strings should be handled, you could do `haupt_person[:t_Nachname].presence || 'unbekannt'` here (it's similar to `present?` and returns `nil` for blank strings)

lib/access_import/haupt_person.rb
toupeira: You could use `#find` instead

lib/access_import/access_import.rb
```diff
@@ -0,0 +1,34 @@
+require_all 'lib/access_import'
```
casaper: So you mean if AccessImport'd be a module instead of a class it would auto load it?

db/migrate/20170727144624_add_import_store_to_tables.rb
```diff
@@ -0,0 +1,7 @@
+class AddImportStoreToTables < ActiveRecord::Migration[5.1]
+  def change
+    add_column :clients, :access_import, :jsonb
+    add_column :volunteers, :access_import, :jsonb
+    add_column :journals, :access_import, :jsonb
```
casaper: But wouldn't this mean to change things inside the models?

lib/access_import/acc_utils.rb
```diff
@@ -0,0 +1,54 @@
...
+  def parse_date_fields(record, keys)
+    keys.each do |key|
+      record[key] = record[key].to_date if record[key]
+    end
+    record
+  end
+
+  def salutation(anrede)
+    return 'mrs' if anrede == 'Frau'
+    return 'mr' if anrede == 'Herr'
```
casaper: no, that is the value that is going to the db. So it should look like those Symbols in the model.

lib/access_import/client_transform.rb
casaper: but nil is not going to be validated by the model, as far as I can see. no?

lib/access_import/acc_utils.rb
```diff
@@ -0,0 +1,54 @@
...
+    keys.each do |key|
+      record[key] = record[key].to_datetime if record[key]
+    end
+    record
+  end
+
+  def parse_date_fields(record, keys)
+    keys.each do |key|
+      record[key] = record[key].to_date if record[key]
+    end
```
casaper: but how do I gain control on what keys are transformed with  `transform_values`? 

Because I'm not transforming all values in the hash, and that is definitely not the desired beaviour here.

db/migrate/20170727144624_add_import_store_to_tables.rb
casaper: And doesn't this even require creation of an extra model?
Plus the columns can be dropped again, after import is run for the very last time.

lib/access_import/acc_utils.rb
nuritnt: yes I understand that when you implemented this, this was still included, but I don't agree with B. Implementing code for the sake of "maybe we could use it in the future" is in my opinion not optimal and this causes again more inconsistency. 
Removing, adding back, removing or refactoring it one more time is just a back and forth ;)
(And it was an acceptance criteria/requested from them to remove the gender)

lib/access_import/acc_utils.rb
casaper: Or maybe with passing the hash values with select and then do `transform_values` and afterwards merge the transformed key's back into the original hash?

casaper: But that wouldn't look any nicer to me, to be honest 😕

lib/access_import/client_transform.rb
casaper: the unbekannt is there, to calm our validations...

lib/access_import/acc_utils.rb
casaper: This is an import, it is not for production use, and it does and should not be seen as part of our application. And advancing to map Assignments from the Access model, could very well include the need of a gender field.

lib/tasks/access.rake
```diff
@@ -0,0 +1,9 @@
+require_all 'lib/access_import'
+
+namespace :access do
+  desc 'TODO'
+  task import: :environment do
+    @acimport = AccessImport.new('db/volunteers_data.accdb')
```
kitsane: Should `*.accdb` files be `.gitignore`d?

casaper: yes. definiteley! Not only for keeping git clean, this data is also private, so should never ever be in our git repo.

casaper:
@toupeira 
> Just some smaller comments, as always some tests would be great ;-)

I would totally like to have tests. but how. I would recommend to have this as an own story, because if it is as I expect, not very trivial to test.

lib/access_import/client_transform.rb
```diff
@@ -0,0 +1,127 @@
...
+      return [begleitet.first[1], begleitete.except(begleitet.first[0])]
+    end
+    begleitet = begleitete.select do |_key, beg|
+      haupt_person[:t_Vorname].include? beg[:t_Vorname]
+    end
+    if begleitet.size.positive?
+      return [begleitet.first[1], begleitete.except(begleitet.first[0])]
+    end
+    # nothing worked, just take first
+    [begleitet.first[1], begleitete.except(begleitet.first[0])]
```
chrysalag: Since this is the same with the return of the last if, is the last if needed?

db/migrate/20170727144624_add_import_store_to_tables.rb
casaper: @toupeira 
Did you have the way I did it now in mind?

casaper:
@toupeira How can I autoload the other classes AccessImport uses?

PR #138, opened 2017-07-27
language preselection

chrysalag:
# [Story in Trello](https://trello.com/c/1M3b5GvQ/64-language-preselection)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/language_skill.rb
+++ b/app/models/language_skill.rb
@@ -5,6 +5,8 @@ class LanguageSkill < ApplicationRecord
 
   def language_name
     return '' if language.blank?
+    return 'Dari' if language == 'DR'
+    return 'Farsi' if language == 'FS'
     I18nData.languages(I18n.locale)[language]
   end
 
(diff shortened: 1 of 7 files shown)


app/models/language_skill.rb
```diff
@@ -5,6 +5,8 @@ class LanguageSkill < ApplicationRecord
 
   def language_name
     return '' if language.blank?
+    return 'Dari' if language == 'DR'
+    return 'Farsi' if language == 'FS'
```
toupeira: It would be great if we could add custom languages to I18nData so we don't have to duplicate it here, but the gem doesn't seem to support that.

app/views/assignments/_language_skills_language_select.html.slim
```diff
@@ -7,5 +7,7 @@
   ul.dropdown-menu
     li
       = link_to t('all'), request_params_filter({language_skills_language_cont: ''})
+    - Client.first_languages.each do |s|
+      li= link_to s[0], request_params_filter({language_skills_language_cont: s[1]})
```
toupeira: You can also do `.each do |(name, code)|` to deconstruct the array items into their own variables.

PR #161, opened 2017-08-04
get pagination translated and with bootstrap elements in view

casaper:
Error with missing translation was:

```
I18n::MissingTranslationData at /clients
translation missing: de.will_paginate.page_gap
```

- both from gems
- volunteers also have pagination now

# [Story in Trello](#)

### Done for approval?

* [ ] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [ ] Translated
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/views/clients/index.html.slim
+++ b/app/views/clients/index.html.slim
@@ -8,9 +8,6 @@
     .col-xs-4
       = button_link t('need_accompanying'), need_accompanying_clients_url
 
-.text-center
-  = will_paginate @clients, renderer: BootstrapPagination::Rails
-
 = render 'columns', clients: @clients
 
 .row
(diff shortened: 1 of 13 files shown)


chrysalag:
Also, we don't use pagination only in index views but also in matching ones. So changes should also be done there.

app/views/clients/index.html.slim
```diff
@@ -1,4 +1,4 @@
-= will_paginate
+= will_paginate @clients, renderer: BootstrapPagination::Rails
```
chrysalag: maybe with the new gem is needed, but the will_paginate doesn't require the name of the collection if one is in the same controller. That's why it was not included here but included in the other will_paginate calls.

casaper: I don't know. I was just following (C&P) the instructions on [bootstrap-ruby/will_paginate-bootstrap](https://github.com/bootstrap-ruby/will_paginate-bootstrap)
Could be that this could also be skipped.

PR #162, opened 2017-08-04
Fix to imports: Volunteers and Clients get status from t_RollenEnde plus bugfix with language_skills

casaper:
# [Story in Trello](https://trello.com/c/f3bB3jDu/102-import-client-volunteer-use-access-creation-time-and-updated-time-for-createdat-and-updatedat)

* Languages where not imported correct
  * foreign key mixup caused wrong language to be joined
  * levels where mapped wrong
* creation time and update time are taken from the access db and used on the imported client and volunteer records.
  * this allows to have usable sorting and more. I recon it would simply be terrible without
* Records that have a RollenEnde date present get RESIGNED (Volunteer) or FINISHED (Client) state
* Volunteers have a first_language collumn, that just displayed the first language in the list so far. That cannot be the propper implementation, because if there is multiple language_skills for a record, the chances are 50/50 at best, that this really is the natively spoken first language 😉 

### Done for approval?

* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/concerns/language_and_schedule_references.rb
+++ b/app/models/concerns/language_and_schedule_references.rb
@@ -7,10 +7,7 @@ module LanguageAndScheduleReferences
 
     delegate :native_languages, to: :language_skills
     delegate :foreign_languages, to: :language_skills
-
-    def native_language
-      native_languages.first || LanguageSkill.new
-    end
+    delegate :native_language, to: :language_skills
 
     has_many :schedules, as: :scheduleable, dependent: :destroy
     accepts_nested_attributes_for :schedules
--- a/app/models/concerns/language_and_schedule_references.rb
+++ b/app/models/concerns/language_and_schedule_references.rb
@@ -5,6 +5,13 @@ module LanguageAndScheduleReferences
     has_many :language_skills, as: :languageable, dependent: :destroy
     accepts_nested_attributes_for :language_skills, allow_destroy: true
 
+    delegate :native_languages, to: :language_skills
+    delegate :foreign_languages, to: :language_skills
+
+    def native_language
+      native_languages.first || LanguageSkill.new
+    end
+
     has_many :schedules, as: :scheduleable, dependent: :destroy
     accepts_nested_attributes_for :schedules
 
--- a/app/models/language_skill.rb
+++ b/app/models/language_skill.rb
@@ -1,13 +1,20 @@
 class LanguageSkill < ApplicationRecord
   belongs_to :languageable, polymorphic: true, optional: true
 
+  LANGUAGE_LEVELS = [:native_speaker, :fluent, :good, :basic].freeze
+
   scope :native_languages, lambda {
     where(level: 'native_speaker')
       .order("CASE WHEN language = 'DE' THEN 1 WHEN language != 'DE' THEN 2 END")
   }
-  scope :foreign_languages, -> { where.not(id: native_languages.first&.id) }
+  scope :foreign_languages, lambda {
+    return none unless native_language.id
+    where.not(id: native_language.id)
+  }
 
-  LANGUAGE_LEVELS = [:native_speaker, :fluent, :good, :basic].freeze
+  def self.native_language
+    native_languages.first || LanguageSkill.new
+  end
 
   def language_name
     return '' if language.blank?
--- a/app/models/language_skill.rb
+++ b/app/models/language_skill.rb
@@ -1,6 +1,12 @@
 class LanguageSkill < ApplicationRecord
   belongs_to :languageable, polymorphic: true, optional: true
 
+  scope :native_languages, lambda {
+    where(level: 'native_speaker')
+      .order("CASE WHEN language = 'DE' THEN 1 WHEN language != 'DE' THEN 2 END")
+  }
+  scope :foreign_languages, -> { where.not(id: native_languages.first&.id) }
+
   LANGUAGE_LEVELS = [:native_speaker, :fluent, :good, :basic].freeze
 
   def language_name
--- a/app/views/volunteers/_volunteer.html.slim
+++ b/app/views/volunteers/_volunteer.html.slim
@@ -6,8 +6,8 @@ tr
   td = volunteer.contact.postal_code
   td = volunteer.contact.primary_email
   td = volunteer.birth_year.year if volunteer.birth_year?
-  td = volunteer.language_skills.first.language_name if volunteer.language_skills.any?
-  td = render 'language_skills/show_inline', language_skills: volunteer.language_skills
+  td = volunteer.native_language&.language_name
+  td = render 'language_skills/show_inline', language_skills: volunteer.foreign_languages
   td = volunteer.education
   td = volunteer.profession
   td #{volunteer.working_percent}%
--- a/lib/access_import/access_import.rb
+++ b/lib/access_import/access_import.rb
@@ -52,10 +52,11 @@ class AccessImport
   end
 
   def instantiate_all_accessors
-    Dir['lib/access_import/accessors/*.rb'].map do |file|
-      File.basename(file, '.*').camelize.constantize.new(@acdb) unless file.include? 'accessor.rb'
-    end.compact
-  end
+    Dir['lib/access_import/accessors/*.rb']
+      .reject { |file| File.basename(file, '.*') == 'accessor' }.map do |file|
+        File.basename(file, '.*').camelize.constantize.new(@acdb)
+      end
+  endFile.basename(file, '.*') == 'accessor'
 
   def make_class_variables(*accessors)
     accessors.each do |accessor|
--- a/lib/access_import/access_import.rb
+++ b/lib/access_import/access_import.rb
@@ -10,39 +10,46 @@ class AccessImport
   end
 
   def make_clients
-    client_transformer = ClientTransform.new(@begleitete, @haupt_person, @familien_rollen)
-    client_count_before = Client.count
-    @personen_rolle.all_clients.each do |key, ac_client|
-      next if Import.where(
-        importable_type: 'Client', access_id: ac_client[:pk_PersonenRolle].to_i
-      ).any?
-      client_attrs = client_transformer.prepare_attributes(ac_client)
-      client = Client.new(client_attrs)
-      client.created_at = ac_client[:d_Rollenbeginn]
-      client.updated_at = ac_client[:d_MutDatum]
+    transformer = ClientTransform.new(@begleitete, @haupt_person, @familien_rollen)
+    make(@personen_rolle.all_clients, transformer, Client) do |client, personen_rolle|
+      client = personen_rollen_create_update_conversion(client, personen_rolle)
       client.user = User.first
-      puts "Importing personen_rolle #{key} to Client.id: #{client.id}" if client.save!
+      if personen_rolle[:d_Rollenende]
+        client.state = Client::FINISHED
+        " -- Client was set to state finished at #{personen_rolle[:d_Rollenende]}  ----"
+      end
     end
-    puts "Imported #{Client.count - client_count_before} new clients from MS Access Database."
   end
 
   def make_volunteers
-    volunteer_transformer = VolunteerTransform.new(@haupt_person)
-    count_before_import = Volunteer.count
-    @personen_rolle.all_volunteers.each do |key, ac_volunteer|
-      next if Import.where(
-        importable_type: 'Volunteer', access_id: ac_volunteer[:pk_PersonenRolle].to_i
-      ).any?
-      volunteer_attrs = volunteer_transformer.prepare_attributes(ac_volunteer)
-      volunteer = Volunteer.new(volunteer_attrs)
-      volunteer.created_at = ac_volunteer[:d_Rollenbeginn]
-      volunteer.updated_at = ac_volunteer[:d_MutDatum]
-      puts "Importing personen_rolle #{key} to Volunteer.id: #{volunteer.id}" if volunteer.save!
+    transformer = VolunteerTransform.new(@haupt_person)
+    make(@personen_rolle.all_volunteers, transformer, Volunteer) do |volunteer, personen_rolle|
+      volunteer = personen_rollen_create_update_conversion(volunteer, personen_rolle)
+      if personen_rolle[:d_Rollenende]
+        volunteer.state = Volunteer::RESIGNED
+        " -- Volunteer was set to state finished at #{personen_rolle[:d_Rollenende]}  ----"
+      end
     end
-    puts "Imported #{Volunteer.count - count_before_import} new volunteers from MS Access Datbase."
   end
 
-  def class_accessor(*accessors)
+  def personen_rollen_create_update_conversion(model_record, personen_rolle)
+    model_record.created_at = personen_rolle[:d_Rollenbeginn]
+    model_record.updated_at = personen_rolle[:d_MutDatum]
+    model_record
+  end
+
+  def make(base_entities, transformer, destination_model)
+    records_before = destination_model.count
+    base_entities.each do |key, entity|
+      next if Import.where(importable_type: destination_model.to_s, access_id: key).any?
+      import_record = destination_model.new(transformer.prepare_attributes(entity))
+      handler_message = yield(import_record, entity)
+      import_record.save!
+      puts "Importing personen_rolle #{key} to #{destination_model}.id: #{import_record.id}#{handler_message}"
+    end
+    message = "Imported #{destination_model.count - records_before} new "
+    puts "#{message}#{destination_model.class.name} from MS Access Database."
+  end
 
   def instantiate_all_accessors
     Dir['lib/access_import/accessors/*.rb']
--- a/lib/access_import/access_import.rb
+++ b/lib/access_import/access_import.rb
@@ -52,11 +52,9 @@ class AccessImport
   end
 
   def instantiate_all_accessors
-    Dir['lib/access_import/accessors/*.rb']
-      .reject { |file| file.slice(-11..-1) == 'accessor.rb' }
-      .map do |file|
-        file.split('/').last.slice(0..-4).camelize.constantize.new(@acdb)
-      end
+    Dir.entries('lib/access_import/accessors').slice(2..-1).map do |file|
+      file.slice(0..-4).camelize.constantize.new(@acdb) unless file == 'accessor.rb'
+    end.compact
   end
 
   def make_class_variables(*accessors)
--- a/lib/access_import/access_import.rb
+++ b/lib/access_import/access_import.rb
@@ -52,8 +52,8 @@ class AccessImport
   end
 
   def instantiate_all_accessors
-    Dir.entries('lib/access_import/accessors').slice(2..-1).map do |file|
-      file.slice(0..-4).camelize.constantize.new(@acdb) unless file == 'accessor.rb'
+    Dir['lib/access_import/accessors/*.rb'].map do |file|
+      File.basename(file, '.*').camelize.constantize.new(@acdb) unless file.include? 'accessor.rb'
     end.compact
   end
 
--- a/lib/access_import/access_import.rb
+++ b/lib/access_import/access_import.rb
@@ -20,6 +20,8 @@ class AccessImport
       ).any?
       client_attrs = client_transformer.prepare_attributes(ac_client)
       client = Client.new(client_attrs)
+      client.created_at = ac_client[:d_Rollenbeginn]
+      client.updated_at = ac_client[:d_MutDatum]
       client.user = User.first
       puts "Importing personen_rolle #{key} to Client.id: #{client.id}" if client.save!
     end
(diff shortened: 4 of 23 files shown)


app/models/concerns/language_and_schedule_references.rb
```diff
@@ -5,6 +5,20 @@ module LanguageAndScheduleReferences
     has_many :language_skills, as: :languageable, dependent: :destroy
     accepts_nested_attributes_for :language_skills, allow_destroy: true
 
+    def first_language
```
koffeinfrei: The name `first_*` implies that the method returns a single item, not a collection.

app/models/concerns/language_and_schedule_references.rb
```diff
@@ -5,6 +5,20 @@ module LanguageAndScheduleReferences
...
+      return [] if language_skills.blank? || language_skills.native.blank?
+      return language_skills.native.german if language_skills.native.german.present?
+      return language_skills.native.limit(1) if language_skills.native.size.positive?
+    end
```
koffeinfrei: Alternative, more readable IMO:

1. leave the guard as is, as it checks the precondition
2. The rest could be done as a simple OR statement

```ruby
return if language_skills.blank? || language_skills.native.blank?

language_skills.native.german.first || language_skills.native.first
```

You could also extract a new scope that returns the first DE or first otherwise.

lib/access_import/access_import.rb
```diff
@@ -20,8 +20,17 @@ def make_clients
       ).any?
       client_attrs = client_transformer.prepare_attributes(ac_client)
       client = Client.new(client_attrs)
+      client.created_at = ac_client[:d_Rollenbeginn]
+      client.updated_at = ac_client[:d_MutDatum]
       client.user = User.first
-      puts "Importing personen_rolle #{key} to Client.id: #{client.id}" if client.save!
+      next unless client.save!
```
koffeinfrei: `save!` will raise an error when saving fails, so you should use `save` here when checking the return value.

lib/access_import/access_import.rb
```diff
@@ -20,8 +20,17 @@ def make_clients
...
+      message = "Importing personen_rolle #{key} to Client.id: #{client.id}"
+      if ac_client[:d_Rollenende]
+        client.state = Client::FINISHED
+        client.save!
```
koffeinfrei: In this case the `client` gets saved twice. Can you reduce this to only one `save` statement? Probably using `valid?` for the above statement and moving `save` after this `if` statement.

lib/access_import/access_import.rb
```diff
@@ -35,7 +44,16 @@ def make_volunteers
...
+      message = "Importing personen_rolle #{key} to Volunteer.id: #{volunteer.id}"
+      if ac_volunteer[:d_Rollenende]
+        volunteer.state = Volunteer::RESIGNED
+        volunteer.save!
+        message += " -- Volunteer was set to resigned at #{ac_volunteer[:d_Rollenende]}  ------"
+      end
+      puts message
     end
     puts "Imported #{Volunteer.count - count_before_import} new volunteers from MS Access Datbase."
   end
```
koffeinfrei: You should be able to DRY up this and extract a common method for ` make_volunteers` and `make_clients`.

lib/access_import/accessors/accessor.rb
```diff
@@ -16,11 +16,17 @@ def add_other_accessors(accessors)
     end
   end
 
+  def add_records_hash
+    @records = hash_all unless @records
```
koffeinfrei: What's the reason for the lazy initialization?

If you need it, the pattern for memoization is more like:

```ruby
def all
  @all ||= hash_all
end

def find(id)
  all[id.to_i]
end

# ... etc.
```

This way you don't need to add `add_records_hash` all over the place (which will likely be forgotten at one point and introduce a bug).

lib/access_import/accessors/sprache_pro_hauptperson.rb
```diff
@@ -26,14 +26,10 @@ def level_keys
...
-
   def where_person(person_id)
-    languages = @records.select do |key|
-      key.to_s == person_id.to_s
-    end
-    languages.map { |s| down_hkeys(s[1]) }
+    add_records_hash
+    @records.select do |_k, val|
+      val[:fk_Hauptperson].to_s == person_id.to_s
+    end.map { |_k, sk| down_hkeys(sk) }
```
koffeinfrei: Please use more explicit variable names, I wouldn't know what `sk` is. Also `_key` is better than `_k`.

app/models/concerns/language_and_schedule_references.rb
casaper: True, there is a naming conflict with rails/ruby naming conventions.
I'll make it a native_language then, if that suits better...

koffeinfrei: > True, there is a naming conflict with rails/ruby naming conventions.

Well it's not ruby or rails specific, the name `first_language` is singular and contains the word `first`, so it clearly implies that the method returns the first item only.

> I'll make it a native_language then, if that suits better...

`native_language **s** (-> plural)

Also, you're calling it as `volunteer.first_language.first`, so the method can just as well directly return 1 item.

app/models/concerns/language_and_schedule_references.rb
casaper: I'm totally aware that this part isn't very elegant at all. But also with your solution, I'm struggling with the problem that I couldn't sort out a scope that relyably only delivered me one item, that isn't or is enclosed in an array.
So I got it crashing either because i was doing `LanguageSkill.first`, or becasue I was returning `[LanguageSkill]` and then the view crashed because of doing `[LanguageSkill].language_name`

lib/access_import/access_import.rb
casaper: But how will I know for sure that save failed?
Because thats something I don't want to happen. It means I'm losing records I'm supposed to import.

lib/access_import/access_import.rb
casaper: So you mean have something seperate, that handles the shell output?

app/models/concerns/language_and_schedule_references.rb
koffeinfrei: There are multiple solutions for the `nil` case:

1. The view checks the return value: `volunteer.first_language.language_name if volunteer.first_language`
1. Use rails' `try`: `volunteer.first_language.try(:language_name)
1. Use ruby's safe operator: `volunteer.first_language&.language_name.
1. Return a valid (but empty) object in any case: `language_skills.native.german.first || language_skills.native.first || LanguageSkill.new`

lib/access_import/accessors/accessor.rb
casaper: I implemented that, because that way I do not need to worry about how many and what Accor class childs I initialize. If their not accessed, it will not slow down the whole thing so much.

> `@all ||= hash_all`

But this will not keep the "hashed" access table stored in the class. So my accessor needs to hash the table every time I'm accessing it.
Some of them are accessed by .all, others use `find(primary_key)`
I'm not sure if that is a good idea in general. But It worked out so far. ..

lib/access_import/access_import.rb
koffeinfrei: Currently the import crashes whenever `save!` is not successful, as it raises an error. The `next unless` doesn't do anything here, as the raised error exits this code.

So either one of the following solutions may apply in this case

1. raise the error and rollback everything when one record fails
1. Check if `client.save` returns `false` and store the failed record in an array and output the failed ones after the whole import process.

lib/access_import/access_import.rb
koffeinfrei: I mean that the 2 methods are very similar and that you should extract the common code to a common, third method.

lib/access_import/accessors/accessor.rb
koffeinfrei: You can store the values on the class:

```ruby
def self.all
  @all ||= hash_all
end
```

When calling from an instance-context call it like `Accessor.all`.

casaper: Thanks. It's much better like that!

lib/access_import/accessors/laender.rb
```diff
@@ -60,10 +60,10 @@ def sanitize_record(rec)
   end
 
   def alpha_2_search(kurzform)
-    ISO3166::Country.all.select { |c| c.alpha2 == kurzform.upcase } || []
+    ISO3166::Country.all.select { |country| country.alpha2 == kurzform.upcase } || []
```
koffeinfrei: `|| []` doesn't do anything, `select` always returns an array.

lib/access_import/accessors/laender.rb
```diff
@@ -60,10 +60,10 @@ def sanitize_record(rec)
...
   end
 
   def alpha_3_search(kurzform)
-    ISO3166::Country.all.select { |c| c.alpha3 == kurzform.upcase } || []
+    ISO3166::Country.all.select { |country| country.alpha3 == kurzform.upcase } || []
```
koffeinfrei: `|| []` doesn't do anything, `select` always returns an array.

lib/access_import/accessors/personen_rolle.rb
```diff
@@ -21,26 +21,26 @@ def map_rolle(z_rolle)
   end
 
   def all_volunteers
-    @records.select do |_id, val|
-      val[:z_Rolle] == 1
+    all.select do |_id, personen_rolle|
+      personen_rolle[:z_Rolle] == 1
```
koffeinfrei: I'd extract those numeric values to a constant (or constants), e.g.

```ruby
ROLES = {
  volunteer: 1,
  # ...
}.freeze
```

```ruby
all.select do |_id, personen_rolle|
  personen_rolle[:z_Rolle] == ROLES.volunteer
end
```

lib/access_import/access_import.rb
```diff
@@ -3,41 +3,71 @@ class AccessImport
...
+    puts "#{message}#{destination_model.class.name} from MS Access Database."
+  end
+
+  def iterate_entitys(base_entitys, transformer, destination_model, import_handler)
+    base_entitys.each do |key, entity|
+      next if Import.where(importable_type: destination_model.to_s, access_id: key).any?
+      record_attributes = transformer.prepare_attributes(entity)
+      import_record = destination_model.new(record_attributes)
+      import_record, handler_message = send(import_handler, import_record, entity)
+      next unless import_record.save!
```
koffeinfrei: `save!` -> `save` (see previous discussion)

lib/access_import/access_import.rb
```diff
@@ -3,41 +3,71 @@ class AccessImport
...
+
+  def make(base_entitys, transformer, destination_model, import_handler)
+    records_before = destination_model.count
+    iterate_entitys(base_entitys, transformer, destination_model, import_handler)
+    message = "Imported #{destination_model.count - records_before} new "
+    puts "#{message}#{destination_model.class.name} from MS Access Database."
+  end
+
+  def iterate_entitys(base_entitys, transformer, destination_model, import_handler)
+    base_entitys.each do |key, entity|
```
koffeinfrei: `base_entitys` -> `base_entities` :policeman:

lib/access_import/access_import.rb
```diff
@@ -3,41 +3,71 @@ class AccessImport
...
-    @personen_rolle.all_clients.each do |key, ac_client|
-      next if Import.where(
-        importable_type: 'Client', access_id: ac_client[:pk_PersonenRolle].to_i
-      ).any?
-      client_attrs = client_transformer.prepare_attributes(ac_client)
-      client = Client.new(client_attrs)
-      client.user = User.first
-      puts "Importing personen_rolle #{key} to Client.id: #{client.id}" if client.save!
+    transformer = ClientTransform.new(@begleitete, @haupt_person, @familien_rollen)
+    make(@personen_rolle.all_clients, transformer, Client, :client_import_handler)
```
koffeinfrei: In ruby you should use blocks for the callback case. There's not need to metaprogram here (although it may be fun :smile:).

app/models/concerns/language_and_schedule_references.rb
```diff
@@ -5,6 +5,18 @@ module LanguageAndScheduleReferences
...
     accepts_nested_attributes_for :language_skills, allow_destroy: true
 
+    def native_language
+      return if language_skills.blank? || language_skills.native.blank?
+      language_skills.native.german.first || language_skills.native.first || LanguageSkill.new
+    end
+
+    def other_languages
+      return language_skills if language_skills.blank? || language_skills.native.blank?
+      language_skills.reject do |lang|
```
casaper: Its not non german, its without the one that is considered native_lanugage. That could well be any other language but german. I prefer german, because I expect to have bilingual people, that have german and some other language as native_spoken in there set. In that case I'd prefer to show german, because it could be misleading otherwise.

app/views/volunteers/_volunteer.html.slim
```diff
@@ -6,8 +6,8 @@ tr
   td = volunteer.contact.postal_code
   td = volunteer.contact.primary_email
   td = volunteer.birth_year.year if volunteer.birth_year?
-  td = volunteer.language_skills.first.language_name if volunteer.language_skills.any?
-  td = render 'language_skills/show_inline', language_skills: volunteer.language_skills
+  td = volunteer.native_language.language_name if volunteer.native_language.present?
```
casaper: there could be no no language_skills records associated at all. They are not a mandatory relation.

lib/access_import/access_import.rb
```diff
@@ -3,41 +3,71 @@ class AccessImport
...
     @acdb = Mdb.open(path)
-    class_accessor(*[Laender, Sprachen, SprachKenntnisse, PersonenRolle, Plz, FamilienRollen,
-                     Begleitete, KostenTraeger, Journale, FallstelleTeilnehmer, Stundenerfassung, Ausbildungen,
-                     FreiwilligenEinsaetze, FreiwilligenEntschaedigung, Kontoangaben].map { |cl| cl.new(@acdb) })
+    class_accessor(
+      *[
+        Laender, Sprachen, SprachKenntnisse, PersonenRolle, Plz, FamilienRollen, Begleitete,
+        KostenTraeger, Journale, FallstelleTeilnehmer, Stundenerfassung, Ausbildungen,
+        FreiwilligenEinsaetze, FreiwilligenEntschaedigung, Kontoangaben
+      ].map { |cl| cl.new(@acdb) }
```
casaper: I do need that array of instances. Its a parameter for `class_accessor(*accessors)`.
class_accors creates attr_reader and class variables from the classes name.

app/models/concerns/language_and_schedule_references.rb
koffeinfrei: Alright, but you should still be able to use `where` instead of `reject`.

lib/access_import/access_import.rb
casaper: So how would I do this in this case? yield? because the `:client_import_handler` is in place, because there is some differences in that block.

app/models/concerns/language_and_schedule_references.rb
koffeinfrei: Hmm, then you could return `LanguageSkill.new` instead of `nil` in `#native_language`. Or use ruby's safe operator.

app/views/volunteers/_volunteer.html.slim
koffeinfrei: Hmm, then you could return LanguageSkill.new instead of nil in #native_language. Or use ruby's safe operator.

lib/access_import/accessors/laender.rb
casaper: makes sense. yea 😉 👍

app/views/volunteers/_volunteer.html.slim
casaper: But maybe i could use the one you suggested before:

`volunteer.native_language.&language_name`

lib/access_import/access_import.rb
koffeinfrei: Yes exactly, instead of `send(import_handler, import_record, entity)` you do `yield import_record, entity` and inline the `#volunteer_import_handler`method as a block.

```ruby
def make_volunteers
  transformer = VolunteerTransform.new(@haupt_person)
  make(@personen_rolle.all_volunteers, transformer, Volunteer) do |volunteer, personen_rolle|
    # ...
  end
end
```

If you would like to keep the `#volunteer_import_handler` method you can also pass the method as a proc argument and call it with `proc.call(..., ...)` instead of yield.

app/views/volunteers/_volunteer.html.slim
koffeinfrei: > But maybe i could use the one you suggested before?

Yes, that's what I meant by "safe operator". The `&` comes before the dot though.

lib/access_import/accessors/personen_rolle.rb
casaper: I wonder how this works?

```ruby
ROLES.volunteer
```
I can only do
```ruby
ROLES[:volunteer]
```

koffeinfrei: Yes sorry, too much JavaScript lately :smile:.

But you could use [OpenStruct](http://ruby-doc.org/stdlib-2.4.1/libdoc/ostruct/rdoc/OpenStruct.html) to achieve this syntax, but it's probably overkill.

casaper: it is overkill. But I still couldn't help myself not using it 😉

app/models/language_skill.rb
```diff
@@ -1,6 +1,11 @@
 class LanguageSkill < ApplicationRecord
   belongs_to :languageable, polymorphic: true, optional: true
 
+  scope :native, -> { where(level: 'native_speaker') }
+  scope :german, -> { where(language: 'DE') }
+  scope :native_language, -> { native.german.first || native.first || LanguageSkill.new }
```
koffeinfrei: This should be a normal method, scopes are expected to return an `ActiveRelation` in order to be chainable.

lib/access_import/access_import.rb
```diff
@@ -3,41 +3,58 @@ class AccessImport
...
+    model_record
+  end
+
+  def make(base_entities, transformer, destination_model)
+    records_before = destination_model.count
+    base_entities.each do |key, entity|
+      next if Import.where(importable_type: destination_model.to_s, access_id: key).any?
+      import_record = destination_model.new(transformer.prepare_attributes(entity))
+      handler_message = yield(import_record, entity)
+      next unless import_record.save
```
koffeinfrei: Maybe you should output a message about the failed import?

lib/access_import/accessors/accessor.rb
```diff
@@ -2,12 +2,11 @@
 
 class Accessor
   include AccUtils
-  attr_reader :records
+  attr_accessor :all
```
koffeinfrei: The reader part will be overwritten by `def all` and the writer part is not used anywhere.

lib/access_import/accessors/personen_rolle.rb
```diff
@@ -8,39 +10,41 @@ def sanitize_record(rec)
...
-    return 'Participant' if z_rolle == 4
-    nil
-  end
+  ACCESS_TO_OWN_ROLES_MAP = [
+    nil,
+    'Volunteer',
+    'Client',
+    'Animator',
+    'Participant'
+  ].freeze
```
koffeinfrei: It probably would be safer to use a hash here, as the index number is a static key. When e.g. someone adds a new entry somewhere in this array a bug will be introduced as the index keys change.

koffeinfrei:
BTW I'm not very fond of having German variable and method names. Is this something you discussed in the scope of this project

casaper:
> BTW I'm not very fond of having German variable and method names. Is this something you discussed in the scope of this project

I'm generally using english variable names, but as our access db is all in german, it is actually better, because it allowes me to differentiate between what is what.
And the argument that this is supposed to be globally usable doesn't hit here, because this is an import for this very specific db .... what is all with german keys and table names.

app/models/language_skill.rb
```diff
@@ -1,6 +1,13 @@
 class LanguageSkill < ApplicationRecord
   belongs_to :languageable, polymorphic: true, optional: true
 
+  scope :native, -> { where(level: 'native_speaker') }
+  scope :native_languages, lambda {
+    return native if native.size <= 1
+    native.order("CASE WHEN language = 'DE' THEN 1 WHEN language != 'DE' THEN 2 END")
+  }
+  scope :non_native_languages, -> { where.not(id: native_languages.first&.id) }
```
koffeinfrei: This may result in the query `where.not(id: nil)` in case `native_languages.first` is `nil`, which will return all records. Is that expected?

koffeinfrei: You could move the method `native_language` to this class and use it for here (`native_language.id`).

app/models/language_skill.rb
```diff
@@ -1,6 +1,13 @@
 class LanguageSkill < ApplicationRecord
   belongs_to :languageable, polymorphic: true, optional: true
 
+  scope :native, -> { where(level: 'native_speaker') }
+  scope :native_languages, lambda {
+    return native if native.size <= 1
```
koffeinfrei: Why is this guard needed? The `order` can be applied in any case.

app/models/language_skill.rb
koffeinfrei: Should this actually only consider the first `native_languages`? Implying from the naming this should return the complementing set to `native_languages`.

app/models/language_skill.rb
koffeinfrei: I'd merge `native` and `native_languages`. Is there a specific reason the two are separate?

Alternatively I'd convert the `native_languages` to an order scope (then you could chain them like `LanguageSkill.native.ordered_by_language`).

PR #163, opened 2017-08-04
Begleitungen import

casaper:
# [Story in Trello](https://trello.com/c/82NCgHNP/54-single-assignement-import)

### Done for approval?

* [ ] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [ ] Translated
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -12,9 +12,10 @@ class Assignment < ApplicationRecord
   has_one :import, as: :importable, dependent: :destroy
   accepts_nested_attributes_for :import, allow_destroy: true
 
-  validates :client_id, uniqueness: { scope: :volunteer_id, message: I18n.t('assignment_exists') }
-
   STATES = [:suggested, :active, :finished, :archived].freeze
+
+  validates :client_id, uniqueness: { scope: :volunteer_id, message: I18n.t('assignment_exists') }
+  validates :state, inclusion: { in: STATES.map(&:to_s) }
 
   scope :no_end, (-> { where(assignment_end: nil) })
   scope :has_end, (-> { where.not(assignment_end: nil) })
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1,11 +1,16 @@
 class Assignment < ApplicationRecord
   belongs_to :client
   accepts_nested_attributes_for :client
+
   belongs_to :volunteer
   accepts_nested_attributes_for :volunteer
+
   belongs_to :creator, class_name: 'User', foreign_key: 'creator_id'
   has_many :hours
   has_many :assignment_journals
+
+  has_one :import, as: :importable, dependent: :destroy
+  accepts_nested_attributes_for :import, allow_destroy: true
 
   validates :client_id, uniqueness: { scope: :volunteer_id, message: I18n.t('assignment_exists') }
 
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -1,3 +1,7 @@
 class Import < ApplicationRecord
   belongs_to :importable, polymorphic: true, optional: true
+
+  def self.get_imported(entity, access_id)
+    find_by(importable_type: entity.to_s.classify, access_id: access_id).importable
+  end
 end
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -2,6 +2,6 @@ class Import < ApplicationRecord
   belongs_to :importable, polymorphic: true, optional: true
 
   def self.get_imported(entity, access_id)
-    find_by(importable_type: entity.to_s.classify, access_id: access_id)&.importable
+    find_by(importable_type: entity.name, access_id: access_id)&.importable
   end
 end
--- a/app/models/import.rb
+++ b/app/models/import.rb
@@ -2,6 +2,6 @@ class Import < ApplicationRecord
   belongs_to :importable, polymorphic: true, optional: true
 
   def self.get_imported(entity, access_id)
-    find_by(importable_type: entity.to_s.classify, access_id: access_id).importable
+    find_by(importable_type: entity.to_s.classify, access_id: access_id)&.importable
   end
 end
--- a/app/models/language_skill.rb
+++ b/app/models/language_skill.rb
@@ -1,12 +1,11 @@
 class LanguageSkill < ApplicationRecord
   belongs_to :languageable, polymorphic: true, optional: true
 
-  scope :native, -> { where(level: 'native_speaker') }
   scope :native_languages, lambda {
-    return native if native.size <= 1
-    native.order("CASE WHEN language = 'DE' THEN 1 WHEN language != 'DE' THEN 2 END")
+    where(level: 'native_speaker')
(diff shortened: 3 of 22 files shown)


app/models/user.rb
```diff
@@ -29,6 +29,7 @@ class User < ApplicationRecord
   validates :role, inclusion: { in: ROLES }
 
   scope :department_assocable, (-> { where(role: CAN_MANAGE_DEPARTMENT) })
+  scope :where_superadmin, (-> { where(role: SUPERADMIN) })
```
toupeira: I would just call this scope `superadmins`

config/locales/de.yml
```diff
@@ -38,6 +38,12 @@ de:
         client: Klient/in
         created_by: Verantwortliche/r
         volunteer: Freiwillige
+        assignment_start: Einsatz beginn
+        assignment_end: Einsatz ende
```
toupeira: Remove spaces

lib/access_import/acc_utils.rb
```diff
@@ -28,6 +28,14 @@ def parse_date_fields(record, *keys)
     record.merge(record.slice(*keys).compact.transform_values(&:to_date))
   end
 
+  def parse_boolean_fields(record, *keys)
+    record.merge(
+      parse_int_fields(record, *keys).slice(*keys).compact.transform_values do |val|
+        val == 1 ? true : false
```
toupeira: You could just do `val == 1` which already returns a boolean

lib/access_import/access_import.rb
```diff
@@ -7,28 +9,53 @@ def initialize(path)
...
-      end
+      client.user_id = @import_user.id
+      client.state = handle_client_state(personen_rolle)
     end
   end
 
+  def handle_client_state(personen_rolle)
+    return Client::FINISHED if personen_rolle[:d_Rollenende]
+    return Client::ACTIVE if personen_rolle[:d_Rollenende].nil?
+    Client::REGISTERED
```
toupeira: I would prefer a multiline `if/elsif` or even `case` but that's just me :)

lib/access_import/access_import.rb
```diff
@@ -7,28 +9,53 @@ def initialize(path)
...
+  def handle_volunteer_state(personen_rolle)
+    return Volunteer::RESIGNED if personen_rolle[:d_Rollenende]
+    return Volunteer::ACTIVE if personen_rolle[:d_Rollenende].nil?
+    Volunteer::REGISTERED
+  end
+
+  def make_assignments
+    transformer = AssignmentTransform.new(@begleitete)
+    @freiwilligen_einsaetze.where_volunteer.each do |key, fw_einsatz|
+      next if Import.where(importable_type: 'Assignment', access_id: key).any?
```
toupeira: You can use `Import.exists?(...)` as a shortcut for `.where(...).any?`

app/models/assignment.rb
```diff
@@ -1,11 +1,17 @@
...
   belongs_to :creator, class_name: 'User', foreign_key: 'creator_id'
 
-  validates :client_id, uniqueness: { scope: :volunteer_id, message: I18n.t('assignment_exists') }
+  has_one :import, as: :importable, dependent: :destroy
+  accepts_nested_attributes_for :import, allow_destroy: true
 
-  STATES = [:suggested, :active].freeze
+  STATES = [:suggested, :active, :finished, :archived].freeze
+
+  validates :client_id, uniqueness: { scope: :volunteer_id, message: I18n.t('assignment_exists') }
```
toupeira: We could/should also add a unique index on the DB level:

```ruby
add_index :assignments, [:client_id, :volunteer_id], unique: true
```

app/models/user.rb
casaper: I named it like that in order to have it clearly distinguishable from User.superadmin? and such. But actually, right...

app/models/assignment.rb
casaper: Not my change actually, because I just moved the line. But never mind. I'll do it.

casaper: Hold on, I need to take that back. I disagree. This is because that uniqueness validation is anyway not an implementation that can stay like this in the future. Because it expects a volunteer and a client to have only one assignment ever, ever and ever. I don't think that it is impossible that a volunteer has an assignment, and when it ends the volunteer has another assignment with the same client.

lib/access_import/acc_utils.rb
casaper: true

lib/access_import/access_import.rb
casaper: I would usually not agree, but I'll be a good boy this time.... 😉

lib/access_import/access_import.rb
casaper: Ah, great. That'll be nicer! thx.

app/models/assignment.rb
chrysalag: It expects that duplicate assignment cannot exist. A volunteer can have many assignments based on our current implementation/validation.

casaper: I'm not mentioning many assignments with different clients. Assignment A first, and then when assignment A is finished another Assignment with the same client again. That is well thinkable, as far as I assume.

casaper: Plus, as the client has:

```ruby
has_one :assignment
```

Could it get more than one then anyway?

toupeira: Ah ok if you were just moving the line around we can leave it as it is.

It's just a general best practice to always enforce uniqueness on the DB level as well, because Rails can't guarantee it completely: https://robots.thoughtbot.com/the-perils-of-uniqueness-validations

app/models/import.rb
```diff
@@ -1,3 +1,7 @@
 class Import < ApplicationRecord
   belongs_to :importable, polymorphic: true, optional: true
+
+  def self.get_imported(entity, access_id)
+    find_by(importable_type: entity.to_s.classify, access_id: access_id)&.importable
```
toupeira: Why not pass the class (`entity`) directly instead of as a symbol?

app/models/language_skill.rb
```diff
@@ -1,6 +1,12 @@
 class LanguageSkill < ApplicationRecord
   belongs_to :languageable, polymorphic: true, optional: true
 
+  scope :native_languages, lambda {
+    where(level: 'native_speaker')
+      .order("CASE WHEN language = 'DE' THEN 1 WHEN language != 'DE' THEN 2 END")
```
toupeira: `CASE WHEN language = 'DE' THEN 1 ELSE 2 END`

app/models/import.rb
casaper: rails will convert that automaticly? Because in the db it will be saved as `'ClassName'`... ?

app/models/language_skill.rb
casaper: Thx. That looks much better.. 👍 😄

app/models/import.rb
casaper: Ah, I know what you mean.

PR #173, opened 2017-08-10
XLS Export of volunteers

casaper:
# Stories in Trello

- [XLS export of volunteer](https://trello.com/c/RsfWXdcZ/1-3-xls-export-of-volunteer)
- [Filter for volunteers applicable for exports](https://trello.com/c/cfzYtbNw/3-3-filter-for-volunteers-applicable-for-exports)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Translated
* [x] Code is reviewed

### Done after the merge?


Gemfile
```diff
@@ -8,6 +8,8 @@ end
 gem 'rails', '~> 5.1'
 
 gem 'autoprefixer-rails'
+gem 'axlsx', '2.1.0.pre'
```
toupeira: Can we relax the version requirement a bit, e.g. `~> 2.1.0`?

Also I always try to add a comment when using a specific version to explain why (e.g. with a link to a relevant bug report or something)

app/controllers/volunteers_controller.rb
```diff
@@ -10,7 +10,15 @@ class VolunteersController < ApplicationController
...
     @q = Volunteer.ransack(params[:q])
-    @volunteers = @q.result.paginate(page: params[:page])
+    respond_to do |format|
+      if params[:format] == 'xlsx'
+        @volunteers = @q.result
+        format.xlsx
+      else
+        @volunteers = @q.result.paginate(page: params[:page])
+        format.html
+      end
```
toupeira: Reformat as:

```ruby
format.xlsx do
  @volunteers = @q.result
end
format.html do
  @volunteers = @q.result.paginate(page: params[:page])
end
```

app/helpers/filter_dropdown_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def filter_link(q_filter, filter_value)
+    filter_value = '' if filter_value == 'all'
+    "#{request.path}?" + {
+      q: @local_q.except(q_filter).merge(q_filter => filter_value)
+    }.to_query
+  end
+
+  def set_local_q
+    @local_q = request.GET[:q] || {}
```
toupeira: - Using an instance variable here makes it more confusing IMHO
- Is there a reason you're using `request.GET` instead of just `params`?
- Maybe rename the method to something like `search_parameters`?
- Maybe move the method to `ApplicationHelper`?

app/helpers/filter_dropdown_helper.rb
```diff
@@ -0,0 +1,89 @@
+module FilterDropdownHelper
+  def list_eq_filter_dropdown(attribute, collection, t_scope: nil)
+    set_local_q
+    t_scope ||= [:simple_form, :options, controller_name.singularize.to_sym]
+    @t_scope = { scope: t_scope.push(attribute) }
```
toupeira: Using an instance variable for this is a bit confusing, it would be better if you could pass the scope on to the other methods.

app/helpers/filter_dropdown_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+      scope: t_scope || [:activerecord, :attributes, controller_name.singularize.to_sym]
+    }
+    q_filters = collection.map { |attribute| "#{attribute}_eq".to_sym }
+    filter_links = q_filters.each_with_index.map do |q_filter, index|
+      list_filter_link q_filter, collection[index].to_sym, boolean_filter_active(@local_q[q_filter])
+    end
+    container_li attribute_group, filter_links, *q_filters
+  end
+
+  def container_li(attribute, filter_links, *q_filters)
```
toupeira: `container_li` seems like a very generic name for this

app/helpers/filter_dropdown_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+    content_tag :li do
+      link_to link_text, filter_link(q_filter, filter_value), class: link_class
+    end
+  end
+
+  def dropdown_menu(filter_links, q_filters)
+    content_tag :ul, class: 'dropdown-menu' do
+      concat deactivate_link(q_filters)
+      concat content_tag :li, '', class: 'divider', role: 'separator'
+      filter_links.collect { |item| concat item }
```
toupeira: Use `.each` here instead of `.collect`

app/helpers/filter_dropdown_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+      q: @local_q.except(*q_filters)
+    }.to_query
+    content_tag :li do
+      link_to t('all'), link_href
+    end
+  end
+
+  def boolean_filter_active(actual_filter)
+    return 'false' if ['false', '', nil].include? actual_filter
+    true if actual_filter == 'true'
```
toupeira: Is it really necessary to use string literals instead of actual booleans?

app/helpers/filter_dropdown_helper.rb
```diff
@@ -0,0 +1,89 @@
...
+  def translate_value(filter_value)
+    return t('all') if ['all', '', nil].include? filter_value
+    t(filter_value, @t_scope)
+  end
+
+  def filter_link(q_filter, filter_value)
+    filter_value = '' if filter_value == 'all'
+    "#{request.path}?" + {
+      q: @local_q.except(q_filter).merge(q_filter => filter_value)
+    }.to_query
```
toupeira: You can probably just do `url_for(q: ...)` so you don't need to worry about `request.path` and `to_query`

test/integration/volunteers_xlsx_export_test.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class VolunteersXlsxExportTest < ActionDispatch::IntegrationTest
+  def setup
+    @superadmin = create :user, role: 'superadmin'
+    10.times { create :volunteer }
+  end
+
+  test 'xlsx file is downloadable' do
+    login_as @superadmin
+    get volunteers_url + '?format=xlsx'
```
toupeira: You can also do `get volunteers_url(format: :xlsx)`

Gemfile
casaper: I simply copy&pasted from their github readme. But It seems to run with no version requirement at all here.

app/helpers/filter_dropdown_helper.rb
casaper: Thats what params[:q] contains. And
```ruby
false != nil
```

So how do I check more elegant for nil than for string 'false' , if `params[:q]` comes with string 'true' and 'false' anyway?

app/helpers/filter_dropdown_helper.rb
casaper: > Is there a reason you're using request.GET instead of just params?

Its too far back, but I remember trying it with params, and then somehow came to request.GET. But I'll check back. Maybe it was just because of beeing stuck with some things in the process.

> Maybe rename the method to something like search_parameters?

Sounds great. That naming isn't great, your right. 👍 😉 

> Maybe move the method to ApplicationHelper?

Yes and no. Yes, because it is generic, no because I think the actual application helper should be cleaned up, meaning split out specific methods, not used globally. I created this  new helper file exactly because I thought that our application helper is starting to get messy.
As all helpers are global in Rails nowadays, ... i thought I start to split ...

casaper: >> Is there a reason you're using request.GET instead of just params?

> Its too far back, ...

`unable to convert unpermitted parameters to hash`

Thats why I did that. What's your oppinion on:

```ruby
params[:q].to_unsafe_hash
```

casaper:
@chrysalag 
As I allready left my scope with creating this buttons, and the redesign of the ui is to be done in a special task, I'll have to ignore this.
This is just a temporary solution, to have my feature at least somewhat useable. A usable interface will be done in its respective story.

app/helpers/filter_dropdown_helper.rb
toupeira: If the string comes from a parameter that's ok, but it's still weird that this method returns either `'false'` as *string* or `true` as *boolean*.

You could just use booleans internally (which will get converted back to strings by `.to_query` in `filter_link`) and reduce this method to `actual_filter == 'true'`, and since it's so short and only used once you could also remove it and just do this check inside `list_boolean_filter_dropdown`.

toupeira: Or if you want to keep using strings for consistency:

```ruby
if actual_filter == 'true'
  'true'
else
  'false'
end
```

app/helpers/filter_dropdown_helper.rb
casaper: Thanks, I wasn't aware of `url_for`. Its obviously multiple times better that way.

app/helpers/filter_dropdown_helper.rb
casaper: Oh jes, that makes allot of sense. My logic was just reverse the logic solution.... 😉

app/controllers/volunteers_controller.rb
```diff
@@ -72,6 +79,7 @@ def set_volunteer
 
   def volunteer_params
     params.require(:volunteer).permit(volunteer_attributes, :rejection_type, :rejection_text,
-      :trial_period, :intro_course, :doc_sent, :bank_account, :evaluation)
+      :trial_period, :intro_course, :doc_sent, :bank_account, :evaluation, :flexible,
```
chrysalag: Why are the availability attributes here? They are already included in the volunteer_attributes through a concern.

casaper: I guess that was was because it wasn't rebased yet. Now everything seems to be in order with those... 👍

PR #174, opened 2017-08-10
Chores/adapt schedule

chrysalag:
# [Story in Trello](https://trello.com/c/inj1vd5w/25-1-adapt-the-new-schedule-from-volunteer-also-to-klient-profile)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Translated
* [x] Code is reviewed

### Done after the merge?


app/helpers/application_helper.rb
```diff
@@ -1,4 +1,8 @@
 module ApplicationHelper
+  def availability_collection
+    [:flexible, :morning, :afternoon, :evening, :workday, :weekend]
```
nuritnt: if we have this as a concern, why do we need it here again? for the translations?

app/models/volunteer.rb
```diff
@@ -82,11 +82,6 @@ def self.state_collection
                           :dancing, :health, :cooking, :excursions, :women, :teenagers,
                           :children, :other_offer].freeze
   REJECTIONS = [:us, :her, :other].freeze
-  AVAILABILITY = [:flexible, :morning, :afternoon, :evening, :workday, :weekend].freeze
-
-  def self.human_boolean(boolean)
-    boolean ? I18n.t('simple_form.yes') : I18n.t('simple_form.no')
```
nuritnt: why did you remove this? 
I thought this was used for example in the checklist which puts `false = no`, `true = yes`

chrysalag: We had this function twice.

app/helpers/application_helper.rb
chrysalag: The concern in a controller concern, used by the controllers and this is a helper method to be used by the views. I am not sure if concerns are supposed to be used somehow by the views.

nuritnt: ah yes totally true :D

PR #182, opened 2017-08-16
volunteer can show their assignments

chrysalag:
# [Story in Trello](https://trello.com/c/QaVwCBJI/37-volunteer-has-access-to-her-assignments)

### Done for approval?

* ~[ ] Story under test~
* [x] Mobile works
* ~[ ] Seeds created~
* ~[ ] Translated~
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -40,7 +40,7 @@ class ApplicationPolicy
   end
 
   def volunteer_related?
-    record.volunteer.user.id == user.id
+    record.volunteer.user_id == user.id
   end
 
   def superadmin_or_record_owner?
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -39,6 +39,10 @@ class ApplicationPolicy
     record.user_id == user.id
   end
 
+  def volunteer_related?
+    record.volunteer.user.id == user.id
+  end
+
   def superadmin_or_record_owner?
     superadmin? || user_owns_record?
   end
@@ -53,6 +57,10 @@ class ApplicationPolicy
 
   def superadmin_or_social_workers_record?
     superadmin? || social_worker? && user_owns_record?
+  end
+
+  def superadmin_or_volunteer_related?
+    superadmin? || volunteer? && volunteer_related?
   end
 
   alias_method :index?,   :deny_all!
--- a/app/views/application/_index_actions.html.slim
+++ b/app/views/application/_index_actions.html.slim
@@ -2,7 +2,7 @@ td= link_to t_action(:show), subject
 td= link_to t_action(:print), { controller: controller_name, action: :show, id: subject.id, print: 'true' }, target: '_blank'
 - if ['clients', 'volunteers'].include? controller_name
   td= link_to t_model(Journal), { controller: 'journals', action: 'index', "#{controller_name.singularize}_id": subject.id }
-td= link_to t_action(:edit), { controller: controller_name, action: :edit, id: subject.id }
-td
-  - if policy(subject.class).destroy?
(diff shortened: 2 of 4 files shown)


app/policies/application_policy.rb
```diff
@@ -39,6 +39,10 @@ def user_owns_record?
     record.user_id == user.id
   end
 
+  def volunteer_related?
+    record.volunteer.user.id == user.id
```
toupeira: Slightly better: `record.volunteer.user_id == user.id` (avoids loading the `user` relation if it's not loaded yet)

app/views/application/_index_actions.html.slim
```diff
@@ -2,7 +2,7 @@ td= link_to t_action(:show), subject
 td= link_to t_action(:print), { controller: controller_name, action: :show, id: subject.id, print: 'true' }, target: '_blank'
 - if ['clients', 'volunteers'].include? controller_name
   td= link_to t_model(Journal), { controller: 'journals', action: 'index', "#{controller_name.singularize}_id": subject.id }
-td= link_to t_action(:edit), { controller: controller_name, action: :edit, id: subject.id }
-td
-  - if policy(subject.class).destroy?
-    = link_to t_action(:destroy), subject, confirm_deleting(subject)
+- if policy(subject.class).index?
+  td= link_to t_action(:edit), { controller: controller_name, action: :edit, id: subject.id }
```
toupeira: I think you can leave off `controller:` here, it will just use the current one by default.

app/views/assignments/_assignment.html.slim
```diff
@@ -5,5 +5,5 @@ tr
   td = t("state.#{assignment.state}", scope: [:simple_form, :options, :assignment]) if assignment.state.present?
   td = assignment.created_at.strftime("%d-%m-%Y")
   td = link_to assignment.creator, assignment.creator if assignment.creator.present?
-  = render 'index_actions', subject: assignment
+  = render 'index_actions', subject: assignment, controller_name: 'assignments'
```
toupeira: Hmm `controller_name` is already a Rails helper, if you do need this variable (i.e. the current controller is something other than `assignments`) you should name it something else to avoid confusion.

app/views/application/_index_actions.html.slim
chrysalag: If I remove the controller: controller_name then when I visit a volunteer's show and try to edit their assignment, it renders the volunteer edit.

app/views/assignments/_assignment.html.slim
chrysalag: This is used only to overwrite the controller_name at the  app/views/application/_index_actions.html.slim template. I need that because otherwise there is the same problem described above.

app/views/application/_index_actions.html.slim
chrysalag: @toupeira If I indeed can remove it somehow, any hint is welcome.

app/views/assignments/_assignment.html.slim
chrysalag: @toupeira If indeed I can avoid the use of controller_name, any hint is welcome.

app/views/application/_index_actions.html.slim
toupeira: Hmm ok I see what you're doing now :)

One option to avoid overwriting `controller_name` would be to use another name for the variable like `target_controller`, and then have a line at the beginning of this view to fallback to `controller_name` so it still works in the cases where you're not overwriting it:

```slim
- target_controller ||= controller_name
```

But you can also leave it as it is now, I'm not sure my suggestion would really be better ;-)

PR #184, opened 2017-08-16
volunteer can add hours

chrysalag:
# [Story in Trello](https://trello.com/c/y83dW1MB/17-3-freiwillige-k%C3%B6nnen-ihre-stunden-erfassen)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Translated
* [x] Code is reviewed

### Done after the merge?


app/controllers/hours_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+    @hours = policy_scope(Hour)
+  end
+
+  def show; end
+
+  def new
+    @hour = Hour.new
+    @assignments = Assignment.where(volunteer: params[:volunteer_id])
+    @assignments_clients = []
+    @assignments.each do |assignment|
```
chrysalag: :+1:

app/helpers/application_helper.rb
```diff
@@ -79,8 +79,8 @@ def multi_conditional_field(f, property, enablers)
       label_html: { class: 'conditional-group' }
   end
 
-  def confirm_deleting(record)
-    if ['assignments', 'journals'].include? controller_name
+  def confirm_deleting(record, controller_name)
```
chrysalag: I pass it down, so as to overwrite it. I might have to find another way to customize the deletion messages.

casaper: But this way you'd need to add a second parameter to every other occasion confirm_deleting is used.
I think something in that manner would be nicer and you don't need to pass controller_name on every other use:
```ruby
def confirm_deleting(record, controller_translate = nil)
  controller_translate ||= controller_name
  if ['assignments', 'journals', 'hours'].include? controller_translate
    confirm_text = t("delete_#{controller_translate.singularize}")
  end
  { method: :delete, data: { confirm: confirm_text || t_confirm_delete(record) } }
end
```

config/routes.rb
```diff
@@ -18,10 +17,12 @@
   resources :volunteers do
     get :seeking_clients, on: :collection
     get :find_client, on: :member, to: 'assignments#find_client'
+    get :volunteer_hours, on: :member
     resources :journals
   end
   resources :volunteer_emails
   resources :profiles, except: [:destroy, :index]
+  resources :hours
```
casaper: I thought this should have been a nested route too, just like journals are!?

And in that case, it would also make that second parameter in confirm_deleting() obsolete.

app/policies/hour_policy.rb
```diff
@@ -0,0 +1,15 @@
+class HourPolicy < ApplicationPolicy
+  class Scope < ApplicationScope
+    def resolve
+      all if superadmin?
```
casaper: And the volunteer shouldn't see her/his own hours?

app/policies/application_policy.rb
```diff
@@ -60,7 +60,7 @@ def superadmin_or_social_workers_record?
   end
 
   def superadmin_or_volunteer_related?
-    superadmin? || volunteer? && volunteer_related?
+    superadmin? || volunteer_related?
```
casaper: Makes sense 👍

app/policies/hour_policy.rb
chrysalag: The volunteer hours are handled with volunteer_hours action and views. That's why the index is available only to the superadmin.

config/routes.rb
chrysalag: That was not specified at the story definition.

app/models/client.rb
```diff
@@ -43,6 +43,10 @@ def to_s
     "#{contact.first_name} #{contact.last_name}"
   end
 
+  def self.having_volunteer
```
casaper:
```ruby
scope :having_volunteer, (-> { where(state: RESERVED) })
```

app/models/volunteer.rb
```diff
@@ -57,6 +59,10 @@ def seeking_clients?
     SEEKING_CLIENTS.include?(state)
   end
 
+  def self.having_clients
```
casaper:
```ruby
scope : having_clients, (-> { where(state: HAVING_CLIENTS) })
```

config/routes.rb
casaper: It was mentioned at the standup, if I'm not wrong

casaper: I think something like that makes more sense:

```ruby
resources :assignments do
  resources :hours
end
resources :volunteers do
  resources :hours
end
```

This brings an automatic scope to either Volunteers hours, or Assignments hours.

app/controllers/hours_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    @hours = policy_scope(Hour)
+  end
+
+  def show; end
+
+  def new
+    @hour = Hour.new(volunteer_id: params[:volunteer_id])
+    @clients = Client.having_volunteer
+    @assignments = Assignment.where(volunteer: params[:volunteer_id])
+    @assignments_clients = if params[:volunteer_id]
```
casaper: That should be a method...

casaper: eg:
```ruby
def clients_select(assignments)
  if params[:volunteer_id]
    @clients_select = assignments.map ......
  else
    @clients_select = assignments.map ......
  end
end
```

app/policies/hour_policy.rb
casaper: If only the superadmin has access to it: What is that scope for then?

db/schema.rb
```diff
@@ -93,6 +93,21 @@
     t.index ["department_id", "user_id"], name: "index_departments_users_on_department_id_and_user_id"
   end
 
+  create_table "hours", force: :cascade do |t|
+    t.date "meeting_date"
+    t.integer "duration"
```
casaper: Why an integer? What if the volunteer wants to record 1.5 Hours?

config/routes.rb
chrysalag: That was for the Assignment Journals, as far as I understood.

casaper: but what is different in this implementation, so that it shouldn't be done like that. We have records related to volunteer and assignments. basically.

app/controllers/hours_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    if current_user.superadmin?
+      redirect_to hours_url, make_notice
+    else
+      redirect_to volunteer_hours_volunteer_path(@hour.volunteer), make_notice
+    end
+  end
+
+  private
+
+  def select_clients
```
casaper: I like that much better 👍 
One thing comes to my mind tho:

How about making the class variable directly in that select_clients method and call this by `after_action :select_clients, only: [:new]`?

This way the action can be:
```ruby
after_action :select_client, only: :new

# ...

def new
  @hour = Hour.new(volunteer_id: params[:volunteer_id])
  authorize @hour
end

# ...

def select_client
   # ...
     @assignments_clients = ...
```

app/controllers/hours_controller.rb
```diff
@@ -0,0 +1,68 @@
+class HoursController < ApplicationController
+  include MakeNotice
+
+  before_action :set_hour, only: [:show, :edit, :update, :destroy]
+  before_action :set_houred
```
toupeira: Not sure I like the word "houred" (verbing weirds language), maybe call it `volunteer` like in the model?

app/views/hours/_form.html.slim
```diff
@@ -0,0 +1,35 @@
...
+    .col-xs-12
+        = f.hidden_field :volunteer_id, value: params[:volunteer_id] || @hour.volunteer.id
+
+  .row
+    .col-xs-12
+      = f.input :meeting_date, include_blank: true, start_year: Date.today.year - 1, end_year: Date.today.year
+
+  .row
+    .col-xs-12.col-md-6
+      = f.input :hours, collection: 1..8
```
toupeira: Maybe move `1..8` and `0..59` into constants, e.g. `Hour::HOUR_RANGE` and `Hour::MINUTE_RANGE`?

config/locales/de.yml
```diff
@@ -292,6 +301,7 @@ de:
       - :month
       - :year
   delete_assignment: Begleitung wirklich löschen?
+  delete_hour: Stunde Eintrag wirklich löschen?
```
toupeira: "Stundeneintrag" looks more correct

config/locales/de.yml
```diff
@@ -292,6 +301,7 @@ de:
...
   delete_journal: Journal Eintrag wirklich löschen?
```
toupeira: And while you're at it: "Journaleintrag", or maybe "Journal-Eintrag"

config/locales/de.yml
```diff
@@ -292,6 +301,7 @@ de:
...
   department_confirm_destroy: Sind Sie sicher das Sie den Standort löschen wollen?
```
toupeira: Also "dass"

db/migrate/20170825090802_add_minutes_to_hours.rb
```diff
@@ -0,0 +1,6 @@
+class AddMinutesToHours < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :hours, :duration, :hours
+    add_column :hours, :minutes, :integer
+  end
+end
```
toupeira: Just a very minor point: if you're introducing a new table in a PR I think it's fine to just update the original migration, instead of adding more migrations to change the table.

app/controllers/hours_controller.rb
toupeira: You can also use the Rails helper `options_from_collection_for_select(@assignments_clients, :id, :to_s)` in the view instead of doing the `.map` here, and then I would just put `@assignments_clients = Assignment.where(volunteer: params[:volunteer_id])` directly into the `#new` method.

db/migrate/20170825090802_add_minutes_to_hours.rb
casaper: Just a side Question, generally to the migrations (not related to this PR): 

Is there a decent (proper) way to squash all those migrations to have it more overseeable?

toupeira: There's the [squasher gem](https://github.com/jalkoby/squasher) but I never used it.

Basically you can take the contents of `db/schema.rb` and use it as a single migration, [GitLab did this once](https://gitlab.com/gitlab-org/gitlab-ce/commit/e8f8abddb0).

casaper: I'll bring this up, so the team and seb can decide if something like that or other ways of slickening the migrations should be done... thx.

toupeira: I wouldn't worry about it until we get to several hundred migrations :)

chrysalag:
@casaper I am now doing the option_from_collection_for_select

app/views/hours/_form.html.slim
```diff
@@ -4,7 +4,8 @@
   .row
     .col-xs-12
       - if action_name ==  'new'
-        = f.association :assignment, collection: @assignments_clients
+        = f.label :assignment
+        = f.select :assignment_id, options_from_collection_for_select(@assignments, 'id', 'client'), {}, class: 'form-control'
```
toupeira: Ah sorry forgot about simple_form :) Apparently you can also [do this](https://github.com/plataformatec/simple_form#associations):

```ruby
f.association :assignment, collection: @assignments_clients, label_method: :client, value_method: :id
```

PR #188, opened 2017-08-18
Performance Reports

casaper:
# [Story in Trello](https://trello.com/c/ORQsT4t6/5-5-kennzahlen-report-for-stadt-z%C3%BCrich)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [ ] Seeds created
* [x] Translated
* [ ] Code is reviewed

### Done after the merge?


app/controllers/performance_reports_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    @performance_report.period_start = @performance_report.period_start.to_date
+    @performance_report.period_end = @performance_report.period_end.to_date
+  end
+
+  def create
+    @performance_report = PerformanceReport.new(report_params.merge(convert_start_end))
+    @performance_report.user = current_user
+    authorize @performance_report
+    if @performance_report.save
+      redirect_to(@performance_report, make_notice)
```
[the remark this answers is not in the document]
casaper: not important, but very true 👍

app/views/performance_reports/_form.html.slim
```diff
@@ -0,0 +1,12 @@
+= simple_form_for(@performance_report) do |f|
+  = simple_error_notice f
+  .row
+    .col-xs-12= f.input :title
+    .col-xs-12= f.input :comment
+
+  .row
+    .col-xs-12= f.input :period, as: :date_range_picker, label: false
+
+    .col-xs-12= f.input :extern
```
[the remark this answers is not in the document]
casaper: I think there is a misinterpretation of the bootstrap grid system. `.row` is only a container for col elements. col-xx-12 elements will be a row. So row isn't needed.

config/locales/en.yml
```diff
@@ -10,6 +10,7 @@ en:
     workday: Workday
   accompaniment_info: The AOZ has several options for volunteering. Find out more about the different areas of volunteering here.
   clear_filters: Clear filters
+  performance_reports: Performantce Reports
```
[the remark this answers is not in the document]
casaper: fixed, thx!

app/controllers/performance_reports_controller.rb
```diff
@@ -0,0 +1,67 @@
+class PerformanceReportsController < ApplicationController
+  include MakeNotice
```
toupeira: Maybe we can move this into `ApplicationController`? It's already used in most controllers anyway.

app/controllers/performance_reports_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+  def show; end
+
+  def new
+    @performance_report = PerformanceReport.new
+    authorize @performance_report
+  end
+
+  def edit
+    @performance_report.period_start = @performance_report.period_start.to_date
+    @performance_report.period_end = @performance_report.period_end.to_date
```
toupeira: Why is `.to_date` necessary here? Maybe we can just use `date` instead of `datetime` in the migration?

app/helpers/application_helper.rb
```diff
@@ -7,6 +7,10 @@ def simple_error_notice(f)
     boostrap_row { f.error_notification } if f.error_notification.present?
   end
 
+  def green_submit(f)
+    bootstrap_row_col { f.button :submit, class: 'btn btn-success' }
+  end
```
toupeira: I would prefer a semantic name, e.g. `success_button` instead of `green_submit`

app/inputs/date_range_picker_input.rb
```diff
@@ -0,0 +1,48 @@
...
+      template.concat range_fields(template)
+      template.concat year_buttons(template)
+    end
+  end
+
+  def year_buttons(template)
+    template.content_tag(:ul, id: "#{object_name}_#{attribute_name}_years",
+      class: 'list-inline year-togglers') do
+      (0..9).to_a.reverse.each do |index|
+        template.concat year_link(template, Time.zone.now.to_date.year - index)
```
toupeira: `.to_date` is unnecessary here

app/models/client.rb
```diff
@@ -35,6 +36,25 @@ class Client < ApplicationRecord
...
+  scope :with_assignment, (-> { joins(:assignment).distinct })
+  scope :with_active_assignment, (-> { with_assignment.merge(Assignment.active) })
+  scope :with_active_assignment_between, lambda { |start_date, end_date|
+    with_assignment.merge(
+      Assignment.active_between(start_date, end_date)
+    )
+  }
+  scope :with_active_assignment_between, lambda { |start_date, end_date|
+    with_assignment.merge(Assignment.active_between(start_date, end_date))
+  }
```
toupeira: Duplicate definition of `:with_active_assignment_between`

app/models/performance_report.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def assignments(zurich = false)
+    assignments = zurich ? Assignment.zurich : Assignment.all
+    {
+      ended: assignments.end_within(period_start..period_end).count,
+      new: assignments.start_within(period_start..period_end).count,
+      active: assignments.active_between(period_start, period_end).count,
+      total: assignments.count
+    }
+  end
```
toupeira: I think all these private methods can be class methods?

app/views/performance_reports/_form.html.slim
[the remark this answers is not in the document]
toupeira: Using `.row` consistently is better though if we change the layout later, and also might work better on mobile.

But it looks like the grid isn't needed here at all, maybe use [`.form-group`](https://getbootstrap.com/docs/4.0/components/forms/) instead?

app/views/performance_reports/index.html.slim
```diff
@@ -0,0 +1,40 @@
...
+        th= t_attr(:period)
+        th= t_attr(:scope)
+        th= t_attr(:extern)
+        th colspan='3'
+
+    tbody
+      - @performance_reports.each do |performance_report|
+        tr
+          td= l(performance_report.created_at.localtime, format: :default)
+          - if performance_report.year.present?
```
toupeira: `performance_report.year?`

app/views/performance_reports/index.html.slim
```diff
@@ -0,0 +1,40 @@
...
+        th= t_attr(:created_at)
+        th= t_attr(:period)
+        th= t_attr(:scope)
+        th= t_attr(:extern)
+        th colspan='3'
+
+    tbody
+      - @performance_reports.each do |performance_report|
+        tr
+          td= l(performance_report.created_at.localtime, format: :default)
```
toupeira: Is `.localtime` necessary here? Maybe we should [set `config.time_zone` instead](http://guides.rubyonrails.org/configuring.html#configuring-rails-components)

app/views/performance_reports/show.html.slim
```diff
@@ -0,0 +1,112 @@
+h1
+  = "#{t_model}"
```
toupeira: `= t_model` :)

app/views/performance_reports/show.html.slim
```diff
@@ -0,0 +1,112 @@
...
+  = ": #{@performance_report.title}" if @performance_report.title.present?
+
+- if @performance_report.comment.present?
```
toupeira: `@performance_report.comment?` etc.

config/routes.rb
```diff
@@ -1,27 +1,29 @@
...
+  # Unauthenticated routes (Generators place them here. Watch out!)
 
-  resources :assignments
   resources :volunteer_applications, only: [:new, :create] do
     get :thanks, on: :collection
   end
 
-  # Authenticated routes start here
+  # Authenticated routes start here (Watch out!)
   devise_for :users
```
toupeira: Hmm the `devise_for` just adds the routes from Devise, it doesn't automagically add authentication to following routes. Or maybe I misunderstand something?

db/migrate/20170818093851_create_performance_reports.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.boolean :extern, default: false
+      t.string :scope
+      t.string :title
+      t.text :comment
+
+      t.timestamps
+    end
+
+    add_column :performance_reports, :deleted_at, :datetime
+    add_index :performance_reports, :deleted_at
```
toupeira: You can move these into `create_table` above:

```ruby
t.datetime :deleted_at, index: true
```

test/factories/contacts.rb
```diff
@@ -2,11 +2,16 @@
   factory :contact do
     sequence :first_name { |n| "first name #{n}" }
     sequence :last_name { |n| "last name #{n}" }
-    street 'Strassenstr. 223'
     extended 'asdfadsf'
-    city 'Zürich'
+    street 'Strassenstr. 223'
+    city 'Rüdisüligen'
```
toupeira: If you ever need random data, the [ffaker gem](https://github.com/ffaker/ffaker) is great for this.

app/views/performance_reports/_form.html.slim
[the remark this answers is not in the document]
casaper: I can't see why nesting every col in a `.row` would make that more mobile friendly?
Actually quite the opposite:
If I'd like to make a responsive layout with bootstrap, eg `.col-xs-12.col-md-6` I should not have multiple `.row`, cos they'd prevent .stacking with following or previous `.col-xs-12.col-md-6`.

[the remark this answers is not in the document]
casaper: But yes, `.form-group` could definiteley be one of the right choices here, but as it was done with collumns everywhere and the scope of my story is far from a frontend redesign, I can't be implementing this here.

app/controllers/performance_reports_controller.rb
casaper: Yes, that has definitely quite something to it. I guess I got here, because of the assumption I made, when I set up the thing. If I had Assignment end and assignment start as datetime, whould an edge-case 31.12.xx 15:12 field be associated with a report period end 31.12.xx ?

casaper: quite possible that that was an unnecessary bloat I made here.

app/models/client.rb
casaper: true

app/models/performance_report.rb
casaper: I'm fairly willing to do so, but what is the advantage/disadvantage?

app/views/performance_reports/index.html.slim
casaper: Year is set with the method called by the ActiveRecord callback that crunches the numbers. If the period_start and period_end cover the whole year it is set with the year.

test/factories/contacts.rb
casaper: I've lost my path last week, with trying to user factory_girl for seeds with faker (*trashed a bit of my unprecious time*)... so I have a faker fake history ... 😉 
But yeah, right.

config/routes.rb
casaper: Thats how I've interpreded it the other day. But I didn't dig deeper. Hope I didn't produce fake news here ;)

app/views/performance_reports/show.html.slim
casaper: 😄

app/controllers/performance_reports_controller.rb
casaper: true

app/models/performance_report.rb
toupeira: Nothing major, just correctness and maybe a bit less confusion :)

config/routes.rb
toupeira: It definitely makes sense to group authorized/unauthorized routes here, but the actual authorization needs to be performed in each controller so the "Watch out!" maybe sends the wrong signal, and we already have `:verify_authorized` watching out for us :)

casaper: I did produce fake news and now where both mixing up things 😉 
`:verify_authorized` is Pundits method, and it doesn't bother about logged in or not (exept, that it requires its @user class variable of course)

My fake news is, that the authentication (login) is triggered by `before_action :authenticate_user!` and in the controllers where authentication is not required it is prevented from beeing triggered by `skip_before_action :authenticate_user!`

So there is no harm or benefit to place these things anywhere in the routes what so ever ... 😉

casaper: But, yes, having the routes in a order that makes sense is definitely an advise I completely want to follow!

PR #190, opened 2017-08-23
Assignment Journals

chrysalag:
# [Story in Trello](https://trello.com/c/wsSM9cNr/18-3-freiwillige-k%C3%B6nnen-ein-inhaltliches-journal-f%C3%BChren)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?


app/helpers/application_helper.rb
```diff
@@ -81,7 +81,7 @@ def multi_conditional_field(f, property, enablers)
 
   def confirm_deleting(record, target_controller = nil)
     target_controller ||= controller_name
-    if ['assignments', 'journals', 'hours'].include? target_controller
+    if ['assignments', 'journals', 'hours', 'assignment_journals'].include? target_controller
       confirm_text = t("delete_#{target_controller.singularize}")
     end
     { method: :delete, data: { confirm: confirm_text || t_confirm_delete(record) } }
```
toupeira: You can also pass a default string to [`I18n.translate`](http://www.rubydoc.info/github/svenfuchs/i18n/master/I18n/Base#translate-instance_method) to avoid listing all the controllers:

```ruby
confirm_text = t("delete_#{target_controller.singularize}", default: t_confirm_delete(record))
{ method: :delete, data: { confirm: confirm_text } }
```

app/models/assignment_journal.rb
```diff
@@ -0,0 +1,5 @@
+class AssignmentJournal < ApplicationRecord
+  belongs_to :volunteer
+  belongs_to :assignment
+  belongs_to :author, class_name: 'User', foreign_key: 'author_id'
```
toupeira: I think the `foreign_key` is not needed here, it will already use the association name to build it.

app/policies/assignment_journal_policy.rb
```diff
@@ -0,0 +1,16 @@
+class AssignmentJournalPolicy < ApplicationPolicy
+  class Scope < ApplicationScope
+    def resolve
+      all if superadmin?
+      resolve_author if volunteer?
```
toupeira: The first line of this method is ignored, add `return` or use a multi-line `if` statement.

app/views/assignment_journals/_form.html.slim
```diff
@@ -0,0 +1,34 @@
+= simple_form_for [@assignment, @assignment_journal] do |f|
+  = simple_error_notice f
+
+  .row
+    .col-xs-12
+      = f.hidden_field :assignment_id, value: params[:assignment_id]
+
+  .row
+    .col-xs-12
+      = f.hidden_field :volunteer_id, value: @assignment.volunteer.id
```
toupeira: Since the hidden fields are invisible I would remove the `.row` and `.col` containers

app/views/assignment_journals/edit.html.slim
```diff
@@ -0,0 +1,7 @@
+.row
+  .col-xs-12
+    h1= t_title(:edit)
+
+== render 'form'
```
toupeira: Single `=` should be fine for `render` (`==` disables HTML escaping which is dangerous/rarely needed)

app/views/assignment_journals/index.html.slim
```diff
@@ -0,0 +1,42 @@
...
+.row
+  .col-xs-12
+    h1= t_title
+
+.row
+  .col-xs-12
+    table.table.table-striped
+      thead
+        tr
+          - if current_user.superadmin?
```
toupeira: Maybe you can check for a specific policy here instead of `superadmin?` directly? (works better if we change the policy later)

app/views/assignment_journals/index.html.slim
```diff
@@ -0,0 +1,42 @@
...
+          th= t_attr(:achievements, AssignmentJournal)
+          th= t_attr(:future, AssignmentJournal)
+          th= t_attr(:comments, AssignmentJournal)
+          th= t_attr(:conversation, AssignmentJournal)
+          th colspan='3'
+
+      tbody
+        - @assignment_journals.each do |record|
+          tr
+            - if current_user.superadmin?
```
toupeira: Same here

app/views/assignment_journals/new.html.slim
```diff
@@ -0,0 +1,7 @@
+.row
+  .col-xs-12
+    h1= t_title(:new)
+
+== render 'form'
```
toupeira: Single `=`

app/views/assignment_journals/show.html.slim
```diff
@@ -0,0 +1,30 @@
+.row
+  .col-xs-12
+    .table-responsive
+      table.table.table-no-border-top
+        tbody
+          - if current_user.superadmin?
```
toupeira: Same here

db/migrate/20170823123608_create_assignment_journals.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.boolean :conversation
+      t.datetime :deleted_at
+      t.belongs_to :volunteer
+      t.belongs_to :assignment
+      t.references :author, index: true, foreign_key: { to_table: :users }
+
+      t.timestamps
+    end
+
+    add_index :assignment_journals, :deleted_at
```
toupeira: You can also use `index: true` above

app/views/assignment_journals/index.html.slim
chrysalag: I have no existing policy for AssignmentJournal that allows only superadmin. Should I create a controller method to authorize only for that?

PR #196, opened 2017-08-30
Certificate for voluntary work

nuritnt:
# [Story in Trello](https://trello.com/c/e3OAMI8a/30-2-dossier-freiwillig-engagiert-kann-manuell-erstellt-werden)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [ ] ~Seeds created~
* [x] Translated
* [ ] Code is reviewed

### Done after the merge?


app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,77 @@
+class CertificatesController < ApplicationController
+  before_action :set_certificate, only: [:show, :destroy, :update, :edit]
+  before_action :set_volunteer, except: [:edit, :show, :destroy]
+
+  def index
+    @certificates = Certificate.where(volunteer: @volunteer)
+    authorize Certificate
```
chrysalag: Shouldn't the authorize come first?

app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def set_volunteer
+    @volunteer = Volunteer.find(params[:volunteer_id])
+  end
+
+  def certificate_params
+    params.require(:certificate).permit(
+      :duration, :duration_end, :duration_start, :hours, :minutes, :text_body, :institution,
+      :funktion, volunteer_contact: [:name, :street, :city], assignment_kinds: Assignment::KINDS
```
chrysalag: NIT picking, but since we use English for naming, shouldn't it be :function?

config/locales/en.yml
```diff
@@ -1,17 +1,118 @@
...
+      last_name: Last name
+      nationality: Nationality
+      phone: Phone number
+      postal_code: Zip
+      profession: Profession
+      salutation: Salutation
+      state: State
+      street: Street
+      updated_at: Updated at
+      zip: Postleitzahl
```
chrysalag: Is there a reason that this is in German?

config/locales/en.yml
```diff
@@ -1,17 +1,118 @@
...
+      uncle: Uncle
+      wife: Wife/Partner
+    role: &id-roles
+      department_manager: Department manager
+      social_worker: Social worker
+      superadmin: Superadmin
+      volunteer: Volunteer
+    salutation: &id-salutations
+      mr: Mr.
+      mrs: Mrs.s
```
chrysalag: typo

app/controllers/certificates_controller.rb
casaper: If you want to bother to translate the whole certificate into entlish, go ahead. But as this Certificate is only available in German, French, and Italian there is somewhat zero sense to this.

config/locales/en.yml
casaper: see above

app/controllers/certificates_controller.rb
chrysalag: I was not talking about the document, of course. We have only German documents. 
I was talking about the attribute `:funktion`.

casaper: As it is specifically for that form, I will not invent a new one. This is not nit picking. Sorry... It is pointless.
It is sort of impossible that this thing is ever used for a other language than german, at least not as is. So the effort to rename db fields, i simply refuse to take here.

app/views/volunteers/show.html.slim
```diff
@@ -12,11 +12,21 @@ nav.navbar.section-navigation
         - if @volunteer.assignments.any?
           li = button_link t_title(:new, BillingExpense), new_volunteer_billing_expense_path(@volunteer)
           li = button_link t_title(:index, BillingExpense), volunteer_billing_expenses_path(@volunteer)
+        - if @volunteer.state == Volunteer::RESIGNED
```
toupeira: If we had used [ActiveRecord::Enum](http://api.rubyonrails.org/classes/ActiveRecord/Enum.html) we could simply do `@volunteer.resigned?` :)

Unfortunately it's a bit tricky to change it now, since it uses integers instead of strings.

app/views/volunteers/show.html.slim
```diff
@@ -12,11 +12,21 @@ nav.navbar.section-navigation
...
+          - elsif @volunteer.certificates.size > 1
+            li= button_link t('.show_certificates'), volunteer_certificates_path(@volunteer)
   ul.list-inline
     - if @volunteer.assignments.any?
-      li = button_link t_title(:new, Hour), new_volunteer_hour_url(@volunteer)
-      li = button_link t_title(:index, Hour), volunteer_hours_path(@volunteer)
-
+      li = button_link t('.report_hours'), new_volunteer_hour_url(@volunteer)
+      li = button_link t('.hour_reports'), volunteer_hours_path(@volunteer)
+    - if current_user.volunteer?
```
toupeira: Maybe reuse a policy here?

app/views/volunteers/show.html.slim
```diff
@@ -158,8 +168,10 @@ h3= t('volunteer_availability')
...
   = render 'assignments/volunteer_index', assignments: @volunteer.assignments
 
-- if current_user.volunteer?
-  = button_link t('edit_profile'), edit_volunteer_path(@volunteer)
-- else
-  = form_navigation_btn :edit
-  = form_navigation_btn :back
+nav.navbar.section-navigation
+  ul.list-inline
+    - if current_user.volunteer?
```
toupeira: Maybe reuse a policy here?

config/locales/de.yml
```diff
@@ -1,61 +1,147 @@
 ---
 de:
+  aaa_injectable: &id-injectables
```
toupeira: Hmm why the `aaa_injectable`? It looks like you're only including it once below `de`.

db/migrate/20170816130953_create_hours.rb
```diff
@@ -2,8 +2,8 @@ class CreateHours < ActiveRecord::Migration[5.1]
   def change
     create_table :hours do |t|
       t.date :meeting_date
-      t.integer :hours
-      t.integer :minutes
+      t.integer :hours, default: 0
+      t.integer :minutes, default: 0
```
toupeira: You should use a new migration to change the default value, since the table already existed for some time.

db/migrate/20170831115230_create_certificate_model.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.date :duration_start
+      t.date :duration_end
+      t.text :institution
+      t.text :text_body
+      t.text :institution
+      t.string :function
+      t.jsonb :volunteer_contact
+      t.jsonb :assignment_kinds
+      t.references :volunteer, foreign_key: true
+      t.references :user, foreign_key: true
```
toupeira: `index: true` also usually makes sense with `#references`

db/migrate/20170905093143_make_boolean_fields_default_false.rb
```diff
@@ -0,0 +1,25 @@
...
+    change_column_default :volunteers, :teenagers, from: nil, to: false
+    change_column_default :volunteers, :children, from: nil, to: false
+    change_column_default :volunteers, :dancing, from: nil, to: false
+    change_column_default :volunteers, :health, from: nil, to: false
+    change_column_default :volunteers, :cooking, from: nil, to: false
+    change_column_default :volunteers, :excursions, from: nil, to: false
+    change_column_default :volunteers, :women, from: nil, to: false
+    change_column_default :volunteers, :unaccompanied, from: nil, to: false
+    change_column_default :volunteers, :zurich, from: nil, to: false
+    change_column_default :volunteers, :other_offer, from: nil, to: false
```
toupeira: Just FYI ActiveSupport has a helper `with_options` which lets you do:

```ruby
with_options from: nil, to: false do |o|
  o.change_column_default :volunteers, :experience
  o.change_column_default :volunteers, :man
  # etc.
end
```

db/migrate/20170905103222_add_anticipated_assignment_type_field.rb
```diff
@@ -0,0 +1,10 @@
+class AddAnticipatedAssignmentTypeField < ActiveRecord::Migration[5.1]
+  def change
+    change_table :assignments do |t|
+      t.string :kind, default: 'accompaniment'
+      t.rename :assignment_start, :period_start
+      t.rename :assignment_end, :period_end
+    end
+    change_column_default :assignments, :state, from: nil, to: 'suggested'
```
toupeira: You can also use `t.change_default` inside `change_table`

test/factories/assignments.rb
```diff
@@ -2,7 +2,9 @@
   factory :assignment do
     client
     volunteer
-    state 'suggested'
+    state ['suggested', 'active'].sample
     association :creator, factory: :user
+    period_start Faker::Date.between(500.days.ago, 200.days.ago)
+    period_end [nil, Faker::Date.between(199.days.ago, 10.days.ago)].sample
```
toupeira: We should have a test for all the factories, there's `FactoryGirl.lint` now which makes this very easy. Not sure why [they recommend putting it into a Rake task though](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#linting-factories), I would just try adding a test which calls `FactoryGirl.lint traits:true`.

app/assets/stylesheets/_utility.scss
```diff
@@ -0,0 +1,48 @@
...
+  margin-right: 30px;
+}
+
+.m-r-20 {
+  margin-right: 20px;
+}
+
+.m-r-10 {
+  margin-right: 10px;
+}
```
toupeira: Not sure how I feel about this, especially since you're only using `.m-b-10` ;-)

Gemfile
```diff
@@ -8,7 +8,7 @@ end
 gem 'rails', '~> 5.1'
 
 gem 'autoprefixer-rails'
-gem 'axlsx'
+gem 'axlsx', '~> 2.1.0.pre'
```
toupeira: When you specify a specific version I would always add a comment with the reason why (e.g. link to bug report), this can save a lot of time when upgrading the app in the future.

app/assets/stylesheets/layout/_content.scss
```diff
@@ -22,30 +22,6 @@
...
-      }
-
-      & + tr td {
-        border-top: 0;
-        padding-top: 25px;
-      }
-    }
-  }
+.row {
+  margin-bottom: 10px;
```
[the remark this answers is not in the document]
casaper: I just untangled and reorganized the whole css. It was globaly effective before. So no change there...

app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  def new
+    @certificate = Certificate.new(volunteer_id: @volunteer.id, user_id: current_user.id)
+    authorize @certificate
+  end
+
+  def create
+    @certificate = Certificate.new(certificate_params.merge(volunteer_id: @volunteer.id,
+      user_id: current_user.id))
+    @certificate.assignment_kinds = assignment_kinds
+    @certificate.text_body = certificate_params['text_body']
```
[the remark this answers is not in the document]
casaper: For unknown reason it did not work when I tryed it before. But I'll try if that was some unknown side effect.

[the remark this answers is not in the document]
casaper: You where right. I guess thats a left over from when I was useing more data inside jsonb fields. Those I couldn't pass with params...

app/controllers/certificates_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def set_certificate
+    @certificate = Certificate.find(params[:id])
+    @volunteer = @certificate.volunteer
+    authorize @certificate
+  end
+
+  def set_volunteer
+    @volunteer = Volunteer.find(params[:volunteer_id])
```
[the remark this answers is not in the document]
casaper: Your right, this wasn't proper. But I think on `:index` and `:new` its necessary.

app/models/certificate.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  belongs_to :volunteer
+  belongs_to :user
+
+  def build_values
+    return unless text_body.nil?
+    update(text_body: default_text_body, function: default_function,
+      assignment_kinds: volunteer.assignment_kinds, volunteer_contact: convert_volunteer_contact,
+      hours: volunteer.hours.sum(:hours) + (volunteer.hours.sum(:minutes) / 60),
+      minutes: volunteer.hours.sum(:minutes) % 60, institution: default_institution)
```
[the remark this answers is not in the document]
casaper: So in that case the calculation is done in ruby instead of in Postgres?

app/models/certificate.rb
```diff
@@ -0,0 +1,55 @@
...
+      TransFair vermittelt Kontakte zwischen Asylsuchenden, Flüchtlingen und Migrant/innen
+      (Begleitete) mit Menschen, die schon länger in der Schweiz leben und sehr gut Deutsch
+      sprechen (Freiwillige). Durch wöchentliche Besuche bei den Begleiteten zuhause während
+      mindestens sechs Monaten oder im Gruppenunterricht fördern die Freiwilligen gezielt und konkret die
+      soziale und Berufliche Integration der Begleiteten. Dadurch ergänzen sie die Sozialberatung
+      auf sinvolle Weise. Das von TransFair angebotene Weiterbildungsprogramm (Einführungskurs,
+      Erfahrungsaustausch, Fachveranstaltungen) bietet den Freiwillgen die Gelegenheit sich
+      vertiefter mit Fragen zu Asylverfahren und Integration auseinanderzusetzen.
+
+      ***Individuellen Text hier einfügen***
```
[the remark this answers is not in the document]
casaper: nope... This is a german only document, specifically requested from AOZ. As the pre-printed Template the thing is printed on only has geman, italian and french on it, it is totally pointless to translate this.

app/views/assignments/show.html.slim
```diff
@@ -1,215 +1,171 @@
...
+        = t('assignment_pdf.languages')
+      .col-xs-7.col-input
+        - @assignment.client.language_skills.each do |language|
+          = "#{language.full_language_skills}. "
 
-          li= t('assignment_pdf.agreement_text.list2.item2')
+.row-box
+  .col-xs-12
+      .col-xs-4
+        = t('assignment_pdf.what')
```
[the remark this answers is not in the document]
casaper: I think I've lost a row here anyway

app/views/volunteers/show.html.slim
casaper: The non enum decision was made by @kitsane in early stage of the project. I'm not really sure what his reasons where, so I can't really argue for or against this....

app/views/volunteers/show.html.slim
casaper: I don't see which policy would tell me if it's a volunteer or not. This conditional is only about the text within the button.

The Volunteer should not consider editing himself. 

`t('edit_profile')` = `Profil bearbeiten` 

and the other version for the superadmins is 

`Edit Volunteer`

config/locales/de.yml
casaper: The reason for this is that keys to be injected have always to be defined higher in the yaml file than when there injected with `*id-my-injected-key`
With this I intend to make sure that even sorting the yaml file will ensure those keys to be at the very beginning.

app/views/assignments/show.html.slim
```diff
@@ -1,215 +1,171 @@
...
-
-    .row.row-box
-      .col-xs-12
-        p= t('assignment_pdf.agreement_text.list1.title')
-        ul
-          li= t('assignment_pdf.agreement_text.list1.item1')
+      .col-xs-5
+        '#{t('assignment_pdf.postal_code')} / #{t('assignment_pdf.city')}
+      .col-xs-7.col-input
+        '#{@assignment.volunteer.contact.postal_code} #{@assignment.volunteer.contact.city}
```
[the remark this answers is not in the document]
casaper: done that. thx.

app/views/assignments/show.html.slim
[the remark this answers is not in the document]
casaper: done it. thx

app/views/layouts/certificate.pdf.slim
```diff
@@ -0,0 +1,18 @@
+doctype html
+html.pdf-export-html
+  head
+    meta charset='utf-8'
+    meta http-equiv='X-UA-Compatible' content='IE=edge'
+    meta name='viewport' content='width=device-width, initial-scale=1'
```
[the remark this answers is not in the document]
casaper: done it thx.

app/views/volunteers/show.html.slim
casaper: In the very beginning of the project @kitsane strongly recommended us not to use enum, for whatever reasons he considered. I agree, it would be quite tricky to be changed now...

app/views/volunteers/show.html.slim
casaper: The policy's will only give a true or false in case the user is a superadmin or if the user is volunteer and the record is associated to it. I'd have to create a new policy rule for this, and as this condition is only for what text is beeing displayed for the same link. So actually not a policy matter at all.

app/views/volunteers/show.html.slim
casaper: The policy's will only give a true or false in case the user is a superadmin or if the user is volunteer and the record is associated to it. I'd have to create a new policy rule for this, and as this condition is only for what text is beeing displayed for the same link. So actually not a policy matter at all.

db/migrate/20170831115230_create_certificate_model.rb
casaper: I see no difference in the resulting schema.rb with either 
```ruby
t.references :volunteer, foregin_key: true
t.references :volunteer, foreign_key:true, index: true
t.belongs_to :volunteer, foreign_key: true
```
They all give me the same output in schema.rb:
```ruby
create_table "certificates", force: :cascade do |t|
    t.integer "hours"
    t.integer "minutes"
    t.date "duration_start"
    t.date "duration_end"
    t.text "institution"
    t.text "text_body"
    t.string "function"
    t.jsonb "volunteer_contact"
    t.jsonb "assignment_kinds"
    t.bigint "volunteer_id"
    t.bigint "user_id"
    t.datetime "deleted_at"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.index ["deleted_at"], name: "index_certificates_on_deleted_at"
    t.index ["user_id"], name: "index_certificates_on_user_id"
    t.index ["volunteer_id"], name: "index_certificates_on_volunteer_id"
  end
```

They're all resulting in the same, with an index.
And is it even possible, to create a foreign key in postgres, that doesn't have an index? So far I thought that this is not possible....

app/assets/stylesheets/_utility.scss
casaper: I've hit in to several occations throughout the project where that should have been, or can be used. And keep in mind that Bootstrap 4 actually has such a thing included. So it seems to make sense.

Gemfile
casaper: If I only know that axlsx is failing without and axlsx documentation tells me to do so?

app/assets/stylesheets/pages/_assignment.scss
```diff
@@ -0,0 +1,81 @@
...
+.signature-title-row {
+  @extend .row;
+
+  span {
+    display: block;
+    font-style: italic;
+  }
+}
+.signature-row {
+  @extend .row;
```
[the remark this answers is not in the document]
casaper: Ok, it makes kind of sense. But in this case I'm makeing a class scoped to that specific page and place. So it is not beeing reused anywhere else.
My intention was to reduce class soup (having markup with more classes than elements).

test/factories/assignments.rb
casaper: I'm not sure if that is in the scope of my story...

app/assets/stylesheets/layout/_content.scss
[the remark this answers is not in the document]
casaper: https://github.com/panter/aoz-003/pull/196/files#diff-84f8fd7003c0c9744039eb89c5f85453L20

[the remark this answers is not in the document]
casaper: https://github.com/panter/aoz-003/blob/85596a8bc3ab773d02e929c5d505ca2412d9a9e7/app/assets/stylesheets/layout/_pdf.scss#L20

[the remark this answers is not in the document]
casaper: It may have been a mistake back then, but as it was globally applied there, I would change everything by removeing it now. So thats why I've put it there.

test/factories/assignments.rb
casaper: ... and as there is quite some other things touched, I can write an issue or a story for this, but don't see FactoryTesting as my scope now. sorry...

app/views/volunteers/show.html.slim
casaper: Its a frontend and translation issue. The difference between volunteer or not is just the text in the very same button. Not a policy issue, and in my humble oppinion it shouldn't be turned into that.

app/assets/stylesheets/layout/_content.scss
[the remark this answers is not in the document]
toupeira: Ah ok nevermind, didn't catch that :)

app/views/volunteers/show.html.slim
toupeira: Ok makes sense!

app/assets/stylesheets/application.scss
```diff
@@ -9,4 +11,7 @@
 @import 'layout/content';
 @import 'layout/footer';
 @import 'layout/forms';
-@import 'layout/pdf';
```
chrysalag: Why is that removed?

config/locales/de.yml
toupeira: Ah ok, makes sense.

test/policies/certificate_policy_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_helper'
+
+class CertificatePolicyTest < PolicyAssertions::Test
+  test 'superadmin can use all actions' do
+    assert_permit(create(:user), Certificate, 'new?', 'create?', 'index?', 'show?', 'edit?',
+      'update?', 'destroy?')
+  end
+
+  test 'other have no access' do
```
chrysalag: NIT: others

test/factories/assignments.rb
toupeira: Fair enough, I'll add this myself :)

casaper:
@chrysalag and @toupeira 
Thx for reviews

PR #221, opened 2017-09-12
Import Journals

casaper:
# [Story in Trello](https://trello.com/c/V7AHWKS3/31-2-text-eintr%C3%A4ge-journale-importieren-korrespondenz)

### Done for approval?

* [ ] ~Readme updated if needed~
* [ ] Story under test
* [ ] ~Mobile works~
* [ ] ~Seeds created~
* [ ] ~Translated~
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -22,7 +22,7 @@ PreCommit:
     problem_on_unmodified_line: warn
 
   RailsSchemaUpToDate:
-    enabled: true
+    enabled: false
     problem_on_unmodified_line: warn
 
   RuboCop:
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -22,7 +22,7 @@ PreCommit:
     problem_on_unmodified_line: warn
 
   RailsSchemaUpToDate:
-    enabled: false
+    enabled: true
     problem_on_unmodified_line: warn
 
   RuboCop:
--- a/README.md
+++ b/README.md
@@ -5,17 +5,27 @@
 
 Ruby version: 2.4.1
 
-## Dependencies
+## Table of content
+
+1. [Dependencies](#dependencies)
+1. [Developer Dependencies](#developer-dependencies)
+1. [User seeds for development](#user-seeds-for-development)
+1. [Create initial superadmin account](#create-initial-superadmin-account)
+1. [Sort locale yaml files](#sort-locale-yaml-files)
+1. [Importing from access db with rake task](#importing-from-access-db-with-rake-task)
+1. [LICENSE](#license)
+
+### Dependencies
 
 - Postgresql
 - [ImageMagick](http://www.imagemagick.org/) (for [paperclip](https://github.com/thoughtbot/paperclip))
 
-## Developer Dependencies
+### Developer Dependencies
 
 - [overcommit](https://github.com/brigade/overcommit)
 - [rubocop](https://github.com/bbatsov/rubocop)
 
-## User seeds for development
+### User seeds for development
 
 Use `rails db:seed` to get these users:
 
@@ -29,15 +39,39 @@ Use `rails db:seed` to get these users:
   - email: department_manager@example.com
   - password: asdfasdf
 
-## Sort locale yaml files
+### Create initial superadmin account
+
+Sends a invitaion email to the email address, so that the account can be activated:
+
+```bash
+$  rails setup:superadmin email=email@test.com
+```
+
+Ready to use with standard Password `asdfasdf` (No activation needed):
+
+**Only use this on testing servers that are unable to send emails (_security risk_)**
+
+```bash
+$  rails setup:superadmin_initialized email=email@test.com
+```
+
+### Sort locale yaml files
 
 Run this task, in order to sort the locale files alphabetically.
 
-```shell
-rails i18n:sort
+```bash
+$  rails i18n:sort
 ```
 
-## LICENSE
+### Importing from access db with rake task
+
+Run in the command line:
+
+```bash
+$  rails access:import file=path/to/access_file.accdb
+```
+
+### LICENSE
 
 All the sources created are made available under the terms
 of the GNU Affero General Public License (GNU AGPLv3).
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Use `rails db:seed` to get these users:
 
 ### Create initial superadmin account
 
-Sends a invitaion email to the email address, so that the account can be activated:
+Sends an invitation email to the email address, so that the account can be activated:
 
 ```bash
 $  rails setup:superadmin email=email@test.com
--- a/app/controllers/journals_controller.rb
+++ b/app/controllers/journals_controller.rb
@@ -37,7 +37,7 @@ class JournalsController < ApplicationController
 
   def destroy
     @journal.destroy
-    redirect_to request.referer, make_notice
+    redirect_back(fallback_location: url_for(@journaled))
   end
 
   private
--- a/app/controllers/journals_controller.rb
+++ b/app/controllers/journals_controller.rb
@@ -37,7 +37,7 @@ class JournalsController < ApplicationController
 
   def destroy
     @journal.destroy
-    redirect_to @journaled, make_notice
+    redirect_to request.referer, make_notice
   end
 
   private
--- a/app/models/concerns/deleted_user_relation_fallback.rb
+++ b/app/models/concerns/deleted_user_relation_fallback.rb
@@ -1,9 +1,0 @@
-module DeletedUserRelationFallback
-  extend ActiveSupport::Concern
-
-  included do
-    def user
-      super || User.deleted.find_by(id: user_id)
-    end
-  end
-end
--- a/app/models/concerns/deleted_user_relation_fallback.rb
+++ b/app/models/concerns/deleted_user_relation_fallback.rb
@@ -1,0 +1,9 @@
+module DeletedUserRelationFallback
+  extend ActiveSupport::Concern
+
+  included do
+    def user
+      super || User.deleted.find_by(id: user_id)
+    end
+  end
+end
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -2,13 +2,17 @@ class Journal < ApplicationRecord
   include ImportRelation
 
   belongs_to :user
-  belongs_to :journalable, polymorphic: true, required: false
-
   belongs_to :assignment, required: false
+
+  belongs_to :journalable, polymorphic: true, required: false
 
   default_scope { order(created_at: :desc) }
 
   CATEGORIES = [:telephone, :conversation, :email, :feedback, :file].freeze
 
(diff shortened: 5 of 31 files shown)


.overcommit.yml
```diff
@@ -22,7 +22,7 @@ PreCommit:
     problem_on_unmodified_line: warn
 
   RailsSchemaUpToDate:
-    enabled: true
+    enabled: false
```
chrysalag: Maybe that should go in a separate PR.

lib/access_import/access_import.rb
```diff
@@ -75,6 +76,25 @@ def make_assignments
       assignment.creator_id = @import_user.id
       assignment.save!
     end
+    puts "Imported #{Assignment.count - records_before} new #{Assignment.class.name} "\
+      'from MS Access Database.'
```
toupeira: More readable:

```ruby
puts "Imported %d new %s from MS Access Database." %
  [Assignment.count - records_before, Assignment.name]
```

(you probably meant `Assignment.name`, `Assignment.class.name` is always "Class" ;-))

lib/access_import/journal_transform.rb
```diff
@@ -0,0 +1,35 @@
+require 'acc_utils'
+
+class JournalTransform
+  include AccUtils
+
+  def initialize(*accessors)
+    accessors.each do |accessor|
+      instance_variable_set("@#{accessor.class.name.underscore}", accessor)
+    end
```
toupeira: This doesn't look like it's actually used?

lib/access_import/journal_transform.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def prepare_attributes(journal, person, assignment, user_id)
+    {
+      body: journal[:m_Text],
+      journalable_type: person.class.name,
+      journalable_id: person.id,
```
toupeira: You can probably just do `journalable: person`?

lib/access_import/journal_transform.rb
```diff
@@ -0,0 +1,35 @@
...
+      assignment_id: assignment&.id,
```
toupeira: And `assignment: assignment`

lib/access_import/journal_transform.rb
```diff
@@ -0,0 +1,35 @@
...
+      created_at: journal[:d_ErfDatum],
+      updated_at: journal[:d_MutDatum],
```
toupeira: Not sure if Rails will overwrite these, or how important it is to keep the original dates

lib/access_import/journal_transform.rb
```diff
@@ -0,0 +1,35 @@
...
+    }
+  end
+
+  CATEGORY_MAP = {
+    1 => :telephone,
+    2 => :conversation,
+    3 => :email,
+    4 => :feedback,
+    5 => :file
+  }.freeze
```
toupeira: I would always put constants at the top of the class, or at least before method definitions.

lib/tasks/setup.rake
```diff
@@ -7,4 +7,13 @@ namespace :setup do
       warn 'No email set, run `rails setup:superadmin email=email@test.com`'
     end
   end
+
+  desc 'Create a superadmin with factory'
+  task superadmin_factory: :environment do
+    if ENV['email'].present?
+      FactoryGirl.create(:user, email: ENV['email'], password: 'asdfasdf', role: 'superadmin')
+    else
+      warn 'No email set, run `rails setup:superadmin email=email@test.com`'
```
toupeira: `setup:superadmin_factory`

app/models/journal.rb
```diff
@@ -1,10 +1,18 @@
...
   belongs_to :journalable, polymorphic: true, required: false
 
   default_scope { order(created_at: :desc) }
 
   CATEGORIES = [:telephone, :conversation, :email, :feedback, :file].freeze
 
   validates :category, presence: true
+
+  def user
+    super || User.deleted.find(user_id)
```
toupeira: Maybe use `find_by(id: user_id)` to avoid an exception if the user doesn't exist anymore (to get the same behaviour as `super`)

app/models/journal.rb
```diff
@@ -1,10 +1,18 @@
 class Journal < ApplicationRecord
+  include ImportRelation
+
   belongs_to :user
+  belongs_to :assignment, required: false
```
toupeira: From the docs:

> NOTE: `required` is set to `true` by default and is deprecated. If you don't want to have association presence validated, use `optional: true`.

app/views/journals/index.html.slim
```diff
@@ -16,11 +16,11 @@ h1= t_title(:index, Journal)
...
               td= record.subject
-              td= record.body.truncate(40) if record.body
+              td= record.body&.truncate(40)
               td= link_to record.user.full_name, record.user.profile
-              td= link_to t_action(:show), client_journal_path(@journaled, record)
-              td= link_to t_action(:edit), edit_client_journal_path(@journaled, record)
-              td= link_to t_action(:destroy), client_journal_path(@journaled, record), confirm_deleting(record)
+              td= link_to t_action(:show), params.to_unsafe_hash.merge(action: :show, id: record.id)
+              td= link_to t_action(:edit), params.to_unsafe_hash.merge(action: :edit, id: record.id)
+              td= link_to t_action(:destroy), params.to_unsafe_hash.merge(action: :destroy, id: record.id), confirm_deleting(record)
```
toupeira: Can you try using `url_for(action: :show, id: record.id)` etc. here?

Or actually, I think you can even do `link_to ..., action: :show, id: record.id` without an explicit `url_for` call, though you need to wrap the params in braces for destroy since you're also passing HTML options there.

toupeira:
The whole import code is very complex and could use some tests, but I'm not sure how worthwhile that is since it probably won't be easy :)

lib/access_import/journal_transform.rb
casaper: Right. Ill make it by inheritance. It's the same in all those parameter preparers

app/views/journals/index.html.slim
casaper: The think is, that the partial is used from two different controllers (clients and volunteers), so the partial needs to make two different urls.
I find the generally phantastic Rails documentation quite a bit improvable if it comes to the link_to helper. There is a quadzillion different ways to use the thing.

casaper:
> The whole import code is very complex and could use some tests, but I'm not sure how worthwhile that is since it probably won't be easy :)

I'm aware of this, and I completeley agree.
To test this, I have one Idea in mind, but It would definiteley it's own story, because it'll be complex as well:

The legacy MS Access application has exports that spit out MS Excel files containing eg. All Volunteers, All Active Volunteers, same for Clients and so on. Those could be converted to CSV and then tests could randomly pick rows from them and check if the imported counterpart can be found in our application.

README.md
```diff
@@ -29,15 +39,39 @@ Use `rails db:seed` to get these users:
   - email: department_manager@example.com
   - password: asdfasdf
 
-## Sort locale yaml files
+### Create initial superadmin account
+
+Sends a invitaion email to the email address, so that the account can be activated:
```
toupeira: "a**n** invita**t**ion" 👮

app/controllers/journals_controller.rb
```diff
@@ -37,7 +37,7 @@ def update
 
   def destroy
     @journal.destroy
-    redirect_to @journaled, make_notice
+    redirect_to request.referer, make_notice
```
toupeira: `redirect_to :back` does the same, with this caveat:

> When using redirect_to :back, if there is no referrer, ActionController::RedirectBackError will be raised. You may specify some fallback behavior for this case by rescuing ActionController::RedirectBackError.

Referer might be empty for various reasons (e.g. privacy extensions), it would probably be sufficient to catch `ActionController::RedirectBackError` in `ApplicationController` and redirect to the root as a fallback.

app/models/concerns/deleted_user_relation_fallback.rb
```diff
@@ -0,0 +1,9 @@
+module DeletedUserRelationFallback
+  extend ActiveSupport::Concern
+
+  included do
+    def user
+      super || User.deleted.find_by(id: user_id)
+    end
+  end
+end
```
toupeira: Had a look at the paranoia README to see if there's a more general way to deal with this, one option would be to specify a scope on the association:

> If you want to include associated soft-deleted objects, you can (un)scope the association:
> 
> ``` ruby
> class Person < ActiveRecord::Base
>   belongs_to :group, -> { with_deleted }
> end
> 
> Person.includes(:group).all
> ```

Should be more reliable, and also avoids running another DB query for deleted users.

app/views/journals/_journal_block.html.slim
```diff
@@ -6,7 +6,7 @@
 #journalBlock.collapse
   .row
     .col-xs-12
-      = button_link t('add_journal_entry'), { controller: 'journals', action: 'new', "#{controller_name.singularize}_id": journaled_for.id }, 'success'
+      = button_link t('add_journal_entry'), url_for([journaled_for, Journal]) + '/new', 'success'
```
toupeira: I think `url_for([journaled_for, Journal], action: :new)` should work here as well.

lib/tasks/setup.rake
```diff
@@ -8,12 +8,12 @@ namespace :setup do
     end
   end
 
-  desc 'Create a superadmin with factory'
-  task superadmin_factory: :environment do
+  desc 'Create Superadmin account ready to use (no activation needed)'
+  task superadmin_initialized: :environment do
     if ENV['email'].present?
       FactoryGirl.create(:user, email: ENV['email'], password: 'asdfasdf', role: 'superadmin')
```
toupeira: I think it would be better to just use `User.create` here, FactoryGirl isn't loaded on production and it seems the factory doesn't actually do anything more here (it only sets these three fields as well).

lib/tasks/access.rake
```diff
@@ -1,11 +1,15 @@
...
-    @acimport.make_journal
+    if ENV['file'].present?
+      @acimport = AccessImport.new(ENV['file'])
+      @acimport.make_departments
+      @acimport.make_clients
+      @acimport.make_volunteers
+      @acimport.make_assignments
+      @acimport.make_journal
+    else
+      puts 'No access file set. run "rails access:import file=path/to/access_file.accdb"'
```
toupeira: Use `warn`?

app/controllers/journals_controller.rb
casaper: I used that first, and had exactly that error.

app/views/journals/_journal_block.html.slim
casaper: I tryed that and for some reason I got a `wrong number of arguments (given 2, expected 0..1)` on that. Thats why I ended up like this.

PR #224, opened 2017-09-14
Features/group offers scaffold

chrysalag:
# [Story in Trello](https://trello.com/c/CAC4iLG9/2-3-als-freiwilligenverantwortliche-r-m%C3%B6chte-ich-ein-neues-gruppenangebot-erfassen)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/group_offers_controller.rb
+++ b/app/controllers/group_offers_controller.rb
@@ -47,9 +47,10 @@ class GroupOffersController < ApplicationController
   end
 
   def group_offer_params
-    params.require(:group_offer).permit(:title, :offer_state, :volunteer_state,
+    params.require(:group_offer).permit(:title, :offer_type, :offer_state, :volunteer_state,
       :necessary_volunteers, :volunteer_responsible_state, :description, :women, :men, :children,
       :teenagers, :unaccompanied, :all, :long_term, :regular, :short_term, :workday, :weekend,
-      :morning, :afternoon, :evening, :flexible, :date_time, :department_id, :organization)
+      :morning, :afternoon, :evening, :flexible, :date_time, :department_id, :organization,
+      :location)
   end
 end
--- a/app/controllers/group_offers_controller.rb
+++ b/app/controllers/group_offers_controller.rb
@@ -1,0 +1,56 @@
+class GroupOffersController < ApplicationController
+  before_action :set_group_offer, only: [:show, :edit, :update, :destroy]
+
+  def index
+    authorize GroupOffer
+    @group_offers = policy_scope(GroupOffer)
+  end
+
+  def show; end
+
+  def new
+    @group_offer = GroupOffer.new
+    authorize @group_offer
+  end
+
+  def edit; end
+
+  def create
+    @group_offer = GroupOffer.new(group_offer_params)
+    @group_offer.department = current_user.department.first if current_user.manages_department?
+    authorize @group_offer
+    if @group_offer.save
+      redirect_to @group_offer, make_notice
+    else
+      render :new
+    end
+  end
+
+  def update
+    if @group_offer.update(group_offer_params)
+      redirect_to @group_offer, make_notice
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    @group_offer.destroy
+    redirect_to group_offers_url, make_notice
+  end
+
+  private
+
+  def set_group_offer
+    @group_offer = GroupOffer.find(params[:id])
+    authorize @group_offer
+  end
+
+  def group_offer_params
+    params.require(:group_offer).permit(:title, :offer_state, :volunteer_state,
+      :necessary_volunteers, :volunteer_responsible, :volunteer_responsible_name, :description,
+      :women, :men, :children, :teenagers, :unaccompanied, :all, :long_term, :regular, :short_term,
+      :workday, :weekend, :morning, :afternoon, :evening, :flexible, :date_time, :department_id,
+      :organization)
+  end
+end
--- a/app/controllers/group_offers_controller.rb
+++ b/app/controllers/group_offers_controller.rb
@@ -50,7 +50,7 @@ class GroupOffersController < ApplicationController
     params.require(:group_offer).permit(:title, :offer_type, :offer_state, :volunteer_state,
       :necessary_volunteers, :volunteer_responsible_state, :description, :women, :men, :children,
       :teenagers, :unaccompanied, :all, :long_term, :regular, :short_term, :workday, :weekend,
-      :morning, :afternoon, :evening, :flexible, :date_time, :department_id, :organization,
+      :morning, :afternoon, :evening, :flexible, :schedule_details, :department_id, :organization,
       :location)
   end
 end
--- a/app/controllers/group_offers_controller.rb
+++ b/app/controllers/group_offers_controller.rb
@@ -48,9 +48,8 @@ class GroupOffersController < ApplicationController
 
   def group_offer_params
     params.require(:group_offer).permit(:title, :offer_state, :volunteer_state,
-      :necessary_volunteers, :volunteer_responsible, :volunteer_responsible_name, :description,
-      :women, :men, :children, :teenagers, :unaccompanied, :all, :long_term, :regular, :short_term,
-      :workday, :weekend, :morning, :afternoon, :evening, :flexible, :date_time, :department_id,
-      :organization)
+      :necessary_volunteers, :volunteer_responsible_state, :description, :women, :men, :children,
+      :teenagers, :unaccompanied, :all, :long_term, :regular, :short_term, :workday, :weekend,
+      :morning, :afternoon, :evening, :flexible, :date_time, :department_id, :organization)
   end
 end
--- a/app/models/group_offer.rb
+++ b/app/models/group_offer.rb
@@ -1,39 +1,14 @@
 class GroupOffer < ApplicationRecord
+  TARGET_GROUP = [:women, :men, :children, :teenagers, :unaccompanied, :all].freeze
+  DURATION = [:long_term, :regular, :short_term].freeze
+  OFFER_STATES = [:internal_offer, :external_offer].freeze
+  EXTERNAL_OFFER = 'external_offer'.freeze
+  VOLUNTEER_STATES = [:internal_volunteer, :external_volunteer].freeze
+  VOLUNTEER_RESPONSIBLE_STATES = [:volunteer_accountable, :volunteer_member].freeze
+
   belongs_to :department, optional: true
 
   validates :necessary_volunteers, numericality: { greater_than: 0 }, allow_nil: true
-
-  def self.target_collection
-    [:women, :men, :children, :teenagers, :unaccompanied, :all]
-  end
-
-  def self.duration_collection
-    [:long_term, :regular, :short_term]
-  end
-
-  INTERNAL_OFFER = 'internal_offer'.freeze
-  EXTERNAL_OFFER = 'external_offer'.freeze
-  OFFER_STATES = [INTERNAL_OFFER, EXTERNAL_OFFER].freeze
-
-  INTERNAL_VOLUNTEER = 'internal_volunteer'.freeze
-  EXTERNAL_VOLUNTEER = 'external_volunteer'.freeze
-  VOLUNTEER_STATES = [INTERNAL_VOLUNTEER, EXTERNAL_VOLUNTEER].freeze
-
-  VOLUNTEER_ACCOUNTABLE = 'volunteer_accountable'.freeze
-  VOLUNTEER_MEMBER = 'volunteer_member'.freeze
-  VOLUNTEER_RESPONSIBLE_STATES = [VOLUNTEER_ACCOUNTABLE, VOLUNTEER_MEMBER].freeze
-
-  def self.offer_state_collection
-    OFFER_STATES.map(&:to_sym)
-  end
-
-  def self.volunteer_state_collection
-    VOLUNTEER_STATES.map(&:to_sym)
-  end
-
-  def self.volunteer_responsible_collection
-    VOLUNTEER_RESPONSIBLE_STATES.map(&:to_sym)
-  end
 
   def external?
     offer_state == EXTERNAL_OFFER
--- a/app/models/group_offer.rb
+++ b/app/models/group_offer.rb
@@ -1,0 +1,41 @@
+class GroupOffer < ApplicationRecord
+  belongs_to :department, optional: true
+
+  validates :necessary_volunteers, numericality: { greater_than: 0 }, allow_nil: true
+
+  def self.target_collection
+    [:women, :men, :children, :teenagers, :unaccompanied, :all]
+  end
+
+  def self.duration_collection
+    [:long_term, :regular, :short_term]
+  end
+
+  INTERNAL_OFFER = 'internal_offer'.freeze
+  EXTERNAL_OFFER = 'external_offer'.freeze
+  OFFER_STATES = [INTERNAL_OFFER, EXTERNAL_OFFER].freeze
+
+  INTERNAL_VOLUNTEER = 'internal_volunteer'.freeze
+  EXTERNAL_VOLUNTEER = 'external_volunteer'.freeze
+  VOLUNTEER_STATES = [INTERNAL_VOLUNTEER, EXTERNAL_VOLUNTEER].freeze
+
+  VOLUNTEER_ACCOUNTABLE = 'volunteer_accountable'.freeze
+  VOLUNTEER_MEMBER = 'volunteer_member'.freeze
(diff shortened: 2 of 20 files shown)


db/migrate/20170914094802_create_group_offers.rb
```diff
@@ -0,0 +1,34 @@
+class CreateGroupOffers < ActiveRecord::Migration[5.1]
+  def change
+    create_table :group_offers do |t|
+      t.string :title
+      t.string :offer_state
+      t.string :volunteer_state
+      t.integer :necessary_volunteers
+      t.string :volunteer_responsible
+      t.string :volunteer_responsible_name
```
casaper: Why isn't that solved with relations?

app/models/group_offer.rb
```diff
@@ -0,0 +1,41 @@
+class GroupOffer < ApplicationRecord
+  belongs_to :department, optional: true
+
```
casaper: why isnt there a relation to volunteer?

app/models/group_offer.rb
```diff
@@ -0,0 +1,41 @@
...
+    [:women, :men, :children, :teenagers, :unaccompanied, :all]
+  end
+
+  def self.duration_collection
+    [:long_term, :regular, :short_term]
+  end
+
+  INTERNAL_OFFER = 'internal_offer'.freeze
+  EXTERNAL_OFFER = 'external_offer'.freeze
+  OFFER_STATES = [INTERNAL_OFFER, EXTERNAL_OFFER].freeze
```
casaper: why not just
```ruby
OFFER_STATES  = [:internal_offer, :external_offer].freeze
```

app/models/group_offer.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  INTERNAL_VOLUNTEER = 'internal_volunteer'.freeze
+  EXTERNAL_VOLUNTEER = 'external_volunteer'.freeze
+  VOLUNTEER_STATES = [INTERNAL_VOLUNTEER, EXTERNAL_VOLUNTEER].freeze
```
casaper: why not `VOLUNTEER_STATES = [:internal_volunteer, :external_volunteer].freeze`?

app/models/group_offer.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  VOLUNTEER_ACCOUNTABLE = 'volunteer_accountable'.freeze
+  VOLUNTEER_MEMBER = 'volunteer_member'.freeze
+  VOLUNTEER_RESPONSIBLE_STATES = [VOLUNTEER_ACCOUNTABLE, VOLUNTEER_MEMBER].freeze
```
casaper: same here, why not just an array of symbols?
That would make the later offer_state_collection etc. methods obsolete

casaper: or even do it with enum?

app/models/group_offer.rb
```diff
@@ -0,0 +1,41 @@
...
+  def self.volunteer_state_collection
+    VOLUNTEER_STATES.map(&:to_sym)
+  end
+
+  def self.volunteer_responsible_collection
+    VOLUNTEER_RESPONSIBLE_STATES.map(&:to_sym)
+  end
+
+  def external?
+    offer_state == EXTERNAL_OFFER
```
casaper: a boolean field would make all that easyer

app/models/group_offer.rb
```diff
@@ -0,0 +1,41 @@
...
+  belongs_to :department, optional: true
+
+  validates :necessary_volunteers, numericality: { greater_than: 0 }, allow_nil: true
+
+  def self.target_collection
+    [:women, :men, :children, :teenagers, :unaccompanied, :all]
+  end
+
+  def self.duration_collection
+    [:long_term, :regular, :short_term]
```
casaper: as Constant DURATIONS

app/models/group_offer.rb
```diff
@@ -0,0 +1,41 @@
+class GroupOffer < ApplicationRecord
+  belongs_to :department, optional: true
+
+  validates :necessary_volunteers, numericality: { greater_than: 0 }, allow_nil: true
+
+  def self.target_collection
+    [:women, :men, :children, :teenagers, :unaccompanied, :all]
```
casaper: as constant TARGETS

app/controllers/group_offers_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  def new
+    @group_offer = GroupOffer.new
+    authorize @group_offer
+  end
+
+  def edit; end
+
+  def create
+    @group_offer = GroupOffer.new(group_offer_params)
+    @group_offer.department = current_user.department.first if current_user.manages_department?
```
casaper: and what happens to the superadmin adding Group offers to different departments?
Can superadmins do that now?

chrysalag: The only users that have departments are the department managers. So this line affects only them.

casaper: Ah, now I realize. The association is optional. It makes sense then.

db/migrate/20170914094802_create_group_offers.rb
casaper:
```ruby
has_one :responsible, class_name: 'User', foreign_key: 'responsible_id'
```

casaper: or
```ruby
has_and_belongs_to_many :volunteer
```

app/models/group_offer.rb
toupeira: An array of symbols would also make more sense to me, unless you need to refer to the array items explicitly in the code (which doesn't seem to be the case here).

Also, it's good practice to put constants at the top of a class, after `include`'s/`extend`'s (see [Ruby Style Guide](https://github.com/bbatsov/ruby-style-guide#consistent-classes))

app/policies/group_offer_policy.rb
```diff
@@ -0,0 +1,16 @@
+class GroupOfferPolicy < ApplicationPolicy
+  class Scope < ApplicationScope
+    def resolve
+      return all if superadmin?
+      resolve_department if department_manager?
```
toupeira: Not sure if it's a problem if `Scope#resolve` returns `nil`, alternatively you can use [`none`](http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-none) to return a scope which will never return any records.

app/views/group_offers/_form.html.slim
```diff
@@ -0,0 +1,57 @@
...
+    .col-xs-12
+      = f.input :title
+
+  .row
+    .col-xs-6
+      = f.input :offer_state, collection: GroupOffer.offer_state_collection, as: :radio_buttons
+    .col-xs-6
+      = f.input :volunteer_state, collection: GroupOffer.volunteer_state_collection, as: :radio_buttons
+
+  - if current_user.superadmin?
```
toupeira: Would it make sense to use a policy here?

config/locales/de.yml
```diff
@@ -59,11 +62,21 @@ de:
       zip: Postleitzahl
       hours: Stunden
       minutes: Minuten
+    group_offers:
+      index: &id-help-texts
+        necessary_volunteers: nötig(e) Freiwillige
+        offer_of: Angebot von
+      show: *id-help-texts
+      form:
+        for_external: Organization (Bei externen Gruppenangebot)
```
toupeira: "Organi**s**ation (**b**ei externe**m** Gruppenangebot)"

app/models/group_offer.rb
chrysalag: Because that is part of a [following story](https://trello.com/c/MGYQ5fwy/4-5-als-freiwilligenverantwortliche-r-m%C3%B6chte-ich-ein-oder-mehrere-freiwillige-mit-gruppenangeboten-matchen).

app/models/group_offer.rb
chrysalag: Yes, but according to the provided documents, we need separate checkboxes for internal and external offers. That's why I didn't implement it as a boolean.

db/migrate/20170914094802_create_group_offers.rb
chrysalag: Because that is part of a [following story](https://trello.com/c/MGYQ5fwy/4-5-als-freiwilligenverantwortliche-r-m%C3%B6chte-ich-ein-oder-mehrere-freiwillige-mit-gruppenangeboten-matchen). Maybe I should remove this field because it will be reimplemented anyway in the following story.

app/policies/group_offer_policy.rb
toupeira: This triggers some additional unnecessary queries, why not simply:

```ruby
if superadmin?
  all
elsif department_manager?
  resolve_department
else
  none
end
```

(and also add `delegate :none, to: :scope` in `ApplicationScope`)

app/policies/application_policy.rb
```diff
@@ -19,6 +19,18 @@ def superadmin_or_volunteer?
     superadmin? || volunteer?
   end
 
+  def superadmin_or_department_manager?
+    superadmin? || department_manager?
+  end
+
+  def departments_offer?
+    department_manager? && record.department == user.department.first
```
casaper: What if `user.department.first` doesn't match, but `user.department.last` would?
Or is that impossible?

casaper: Or in Other words, wouldn't it be safer to do:
```ruby
user.department.include? record.department
```
?

app/views/group_offers/index.html.slim
```diff
@@ -0,0 +1,49 @@
+.row
```
casaper: why do you put the h1 into a full width column?

h1 is a full width block element inside container on its own.

app/views/group_offers/edit.html.slim
```diff
@@ -0,0 +1,7 @@
+.row
```
casaper: why do you put the h1 into a full width column?

h1 is a full width block element inside container on its own.

app/views/group_offers/new.html.slim
```diff
@@ -0,0 +1,7 @@
+.row
```
casaper: why do you put the h1 into a full width column?

h1 is a full width block element inside container on its own.

app/views/group_offers/show.html.slim
```diff
@@ -0,0 +1,72 @@
+.row
```
casaper: why do you put the h1 into a full width column?

h1 is a full width block element inside container on its own.

app/views/application/_navigation.html.slim
```diff
@@ -24,6 +24,8 @@ nav.navbar.navbar-top.hidden-print
             li= link_to t_model(Department), department_path(current_user.department.first.id)
           - if policy(Assignment).index?
             li= link_to t('assignments'), assignments_path
+          - if policy(GroupOffer).index?
+            li= link_to t('.group_offers'), group_offers_path
```
casaper: `t('.group_offers')` will return a different translation, depending on the actual controller. In that case it means you'd need to make sure that key .group_offers is available in any `controller_name.action_name.group_offers` key in the yaml files. Very likeley that this will fail.

Id do in the locale yaml:
```yaml
group_offers:
  one: Gruppenangebot
  other: Gruoppenangebote
```

```slim
li= link_to t('group_offers', count: 2), group_offers_path
```
Maybe also not perfect. Id Prefer t to have a Plural option. But unfortunately it doesn't, so that would be a work arround.

config/locales/de.yml
```diff
@@ -59,11 +62,26 @@ de:
       zip: Postleitzahl
       hours: Stunden
       minutes: Minuten
+    group_offers:
+      index: &id-help-texts
+        necessary_volunteers: nötig(e) Freiwillige
```
casaper:
```yaml
necessary_volunteers: Benötigt %{amount} Freiwillige 
```
```ruby
t('.necessary_volunteers', amount: @group_offer.volunteers_needed)
```

config/locales/de.yml
```diff
@@ -59,11 +62,26 @@ de:
...
+        offer_of: Angebot von
+      show: *id-help-texts
+      form:
+        location_external: Ort (bei externem Gruppenangebot)
```
casaper: "Ort (betrifft externe Gruppenangebote)"
?

config/locales/de.yml
```diff
@@ -199,6 +223,28 @@ de:
         department_manager: Standort Betreuer
         name: Name
         user: User
+      group_offer:
+        all: Alle
+        availability: Zeitangaben
+        children: Kinder
+        date_time: Präzise Angaben (Tag und Uhrzeit) und genauen Zeitraum
+        description: Beschreibung des Angebotes
```
casaper: Angebotes -> Angebots

config/locales/de.yml
```diff
@@ -199,6 +223,28 @@ de:
...
+        duration: Dauer
+        location: Ort
+        long_term: Langfristig
+        men: Männer
+        necessary_volunteers: Anzahl nötiger Freiwillige
```
casaper: Anzahl nötiger Freiwillige -> Anzahl der benötigten Freiwilligen

db/migrate/20170914094802_create_group_offers.rb
```diff
@@ -0,0 +1,35 @@
...
+      t.boolean :long_term, default: false
+      t.boolean :regular, default: false
+      t.boolean :short_term, default: false
+      t.boolean :workday, default: false
+      t.boolean :weekend, default: false
+      t.boolean :flexible, default: false
+      t.boolean :morning, default: false
+      t.boolean :afternoon, default: false
+      t.boolean :evening, default: false
+      t.text :date_time
```
casaper: `t.text :date_time` is a misleading column name, as it is the same as t.datetime. I'd rename it to basically something that tells that it is a human text time entry.

test/policies/group_offers_policy_test.rb
```diff
@@ -0,0 +1,25 @@
...
+  test 'social worker and volunteer have no access' do
+    refute_permit(create(:social_worker), GroupOffer, 'new?', 'create?', 'index?', 'show?',
+      'edit?', 'update?', 'destroy?')
+    refute_permit(create(:user_volunteer), GroupOffer, 'new?', 'create?', 'index?', 'show?',
+      'edit?', 'update?', 'destroy?')
+  end
+
+  test 'department manager has limited access' do
+    department_manager = create :department_manager, :with_departments
+    create :group_offer
```
casaper: This way you can never be 100% sure to get the right group offer and department manager.
As create returns the object, you can just do it like following, and it will allways be 100% sure:
```ruby
department_manager = create :department_manager, :with_departments
group_offer_first = create :group_offer
group_offer_second = create :group_offer, department: department_manager.department.last
# ...
refute_permit(department_manager, group_offer_first, 'show?', 'edit?', 'update?')
assert_permit(department_manager, group_offer_second, 'show?', 'edit?', 'update?')
```

test/system/group_offers_test.rb
```diff
@@ -0,0 +1,38 @@
+require 'application_system_test_case'
+
+class GroupOffersTest < ApplicationSystemTestCase
+  def setup
+    create :department
```
casaper: Again `@department = create :department`

test/system/group_offers_test.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  test 'new group offer form' do
+    login_as create(:user)
+    visit new_group_offer_path
+
+    fill_in 'Title', with: 'asdf'
+    page.choose('group_offer_offer_type_internal_offer')
+    page.choose('group_offer_offer_state_full')
+    page.choose('group_offer_volunteer_state_internal_volunteer')
+    select Department.first.contact.last_name, from: 'Department'
```
casaper: `@department.contact.last_name`

config/locales/de.yml
chrysalag: It is stated exactly like that in their document. Also: http://www.crodict.de/nomen/deutsch/Angebot.html

app/views/application/_navigation.html.slim
chrysalag: t('.group_offers') looks for: [en/de].application.navigation.group_offers

app/views/group_offers/edit.html.slim
chrysalag: It adds some space between the title of the form and the rest of form.

app/views/application/_navigation.html.slim
casaper: ok, wasn't aware of that

config/locales/de.yml
casaper: Well, the customer is always right 😉

PR #227, opened 2017-09-15
Refactored the way amount of billing expenses is calculated to...

casaper:
...be calculated automaticly before_validation


app/controllers/billing_expenses_controller.rb
```diff
@@ -55,27 +51,8 @@ def set_volunteer
...
-      100
-    else
-      150
-    end
-  end
-
-  def volunteer_hours_update
-    @volunteer_hours.update(billing_expense_id: @billing_expense.id)
+  def file_name
+    [@volunteer.contact.full_name, @volunteer.hours.maximum(:meeting_date)].join('-').parameterize
```
toupeira: Is the ""Spesenauszahlung-" prefix not needed anymore?

Also for clarity I would call this method `pdf_file_name`

app/models/billing_expense.rb
```diff
@@ -19,4 +21,25 @@ def billing_hours
   def billed_hours
     Hour.where(billing_expense: id)
   end
+
+  def auto_set_amount
+    self.amount = compute_amount
+  end
```
toupeira: Any reason to split this across two methods?

app/models/volunteer.rb
```diff
@@ -70,6 +70,10 @@ class Volunteer < ApplicationRecord
 
   scope :with_hours, (-> { joins(:hours).distinct })
 
+  def billable_hours_sum
+    hours.billable.sum(&:hours) + hours.billable.sum(&:minutes) / 60
+  end
+
```
toupeira: To avoid repeating the logic from `hours_sum` it's also possible to [pass a block to `has_many`](http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many) to define methods which will be available on that association, i.e. something like:

```ruby
has_many ... do
  def total_hours
    sum(&:hours) + sum(&:minutes) / 60
  end
end
```

And then you should be able to do `hours.total_hours` and `hours.billable.total_hours` (or with any other scope). I never used this before so not sure this is really how it works :)

test/factories/billing_expenses.rb
```diff
@@ -1,7 +0,0 @@
-FactoryGirl.define do
-  factory :billing_expense do
-    volunteer
-    user
-    amount 50
-  end
-end
```
toupeira: Why not keep the factory?

app/models/billing_expense.rb
casaper: To avoid writing self.amount 4 times. Basically... no real reason

toupeira: You can also do

```ruby
self.amount =
  if
    ...
  elsif
    ...
  end
```

test/factories/billing_expenses.rb
casaper: Because it was a mere nightmare to get that working and passing FactoryGirl validation.
Thats why i killed it.

casaper: It was not working and its multiple relations caused to either not beeing created for the test or not passing the FactoryGirl validation. If Factories are ment to make ones life easier, this one does certenly not meet that requirement. So it should be dropped.

db/seeds.rb
```diff
@@ -41,25 +41,27 @@ def make_lang_skills
...
-        street: Faker::Address.street_address,
-        primary_email: Faker::Internet.email,
-        primary_phone: Faker::PhoneNumber.phone_number
-      )
-      profile.profession = Faker::Company.profession
-      availability_collection.each do |availability|
-        profile[availability] = [true, false].sample
-      end
+  end
+  user.profile = Profile.new do |profile|
```
chrysalag: Doesn't this just adds a profile only for the last user created? What is the reason for that?

casaper: I don't know why, but it didn't work anymore as it was before, and changeing it this way got it running again.

casaper: For unknown reason the way it was before it didn't work anymore, and changing it this way got it back running again... so ...

casaper:
@nuritnt wrote:
> ... only response is "was already created". wouldn't it be nice, if you would have some kind of feedback ...

Yes, just done that. Makes more sense

test/factories/billing_expenses.rb
toupeira: Fair enough!

app/models/billing_expense.rb
```diff
@@ -11,12 +13,16 @@ class BillingExpense < ApplicationRecord
...
 
-  def billing_hours
-    relevant_hours = Hour.where(billing_expense: id)
-    relevant_hours.sum(&:hours) + relevant_hours.sum(&:minutes) / 60
-  end
-
-  def billed_hours
-    Hour.where(billing_expense: id)
+  def compute_amount
+    hour_count = id ? hours.total_hours : volunteer.hours.billable.total_hours
```
toupeira: Regarding the `id ?`, if you want to check if a record was saved there's also `persisted?` (no need to change it here, just FYI)

PR #232, opened 2017-09-21
user email update

nuritnt:
# [Story in Trello](https://trello.com/c/Sw37xP3x/32-email-change-of-volunteers-does-not-work)

ensuring updating email via the contact is also updated on user emails

### Done for approval?

* [x] Story under test
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -89,4 +89,10 @@ class UserTest < ActiveSupport::TestCase
   test 'automatically assigns primary email on contact' do
     assert_equal @user.profile.contact.primary_email, @user.email
   end
+
+  test 'updating user email contact also on user' do
+    @user.profile.contact.update(primary_email: 'supervolunteer@example.com')
+    @user.reload
+    assert_equal 'supervolunteer@example.com', @user.email
+  end
 end
--- a/test/system/assignments_test.rb
+++ b/test/system/assignments_test.rb
@@ -5,7 +5,7 @@ class AssignmentsTest < ApplicationSystemTestCase
     @user = create :user, email: 'superadmin@example.com'
     Assignment.with_deleted.map(&:really_destroy!)
     @volunteer_user = create :user, role: 'volunteer'
-    @client = create :client, user: @user
+    @client = create :client
     @volunteer = create :volunteer, state: Volunteer::ACTIVE_FURTHER, user: @volunteer_user
   end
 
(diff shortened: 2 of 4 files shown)


test/models/user_test.rb
```diff
@@ -89,4 +89,10 @@ def setup
   test 'automatically assigns primary email on contact' do
     assert_equal @user.profile.contact.primary_email, @user.email
   end
+
+  test 'updating user email contact also on user' do
+    @user.profile.contact.update(primary_email: 'supervolunteer@example.com')
```
chrysalag: `@user` is superadmin. I think you mentioned that the issue was with the volunteers.

nuritnt: I noticed it first for a volunteer, but apparently it concerns all users.

chrysalag: I am sorry, I hadn't noticed.

nuritnt:
This fix apparently it revealed two small "wrong" implemented tests.

test/system/assignments_test.rb
```diff
@@ -22,7 +22,7 @@ class AssignmentsTest < ApplicationSystemTestCase
     assert page.has_text? 'Assignment was successfully created.'
     within '.table-striped' do
       assert page.has_text? 'Suggested'
-      assert page.has_link? 'superadmin@example.com'
+      assert page.has_link? 'superadmin@example.com', @user.email
```
chrysalag: Why not either  `assert page.has_link? @user.email` or `assert page.has_link? 'superadmin@example.com'`?
http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Matchers#has_link%3F-instance_method

nuritnt: This test should assure that the mail of the "creator" is shown at created_by.
With the first implementation we didn't noticed the "wrong" implementation, because we only checked if there is any mail (the superadmin@example.com was assigned to the client, and the test was green, because obviously the mail was displayed) and a client doesn't have the policy to create an assignment anyway. so two-in-one test :)

chrysalag: If you run this test alone, you'll notice that you get a warning: 
`Unused parameters passed to Capybara::Queries::SelectorQuery : ["superadmin@example.com"]`
You cannot check two in one with `.has_link?` method.
The test checks if the email of the creator of the assignment is displayed. So `@user.email` should be sufficient. In the index view we don't display the email of the client. So the test is not mislead. In such a case, I think there would be an error, saying that there are two such links.

PR #235, opened 2017-09-22
Group Offers and Volunteers

chrysalag:
# [Story in Trello](https://trello.com/c/MGYQ5fwy/4-5-als-freiwilligenverantwortliche-r-m%C3%B6chte-ich-ein-oder-mehrere-freiwillige-mit-gruppenangeboten-matchen)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/volunteer.rb
+++ b/app/models/volunteer.rb
@@ -73,6 +73,12 @@ class Volunteer < ApplicationRecord
     joins(:assignments).merge(Assignment.active_between(start_date, end_date)).distinct
   }
   scope :without_assignment, (-> { left_outer_joins(:assignments).where(assignments: { id: nil }) })
+  scope :without_active_assignment, (-> { joins(:assignments).merge(Assignment.ended) })
+  scope :not_responsible, (-> { where.not(id: GroupOffer.pluck(:responsible_id)) })
+  scope :not_in_any_group_offer, lambda {
+    joins('LEFT JOIN group_offers_volunteers ON volunteers.id = group_offers_volunteers.volunteer_id')
+      .where('group_offers_volunteers.volunteer_id IS NULL')
+  }
 
   scope :with_hours, (-> { joins(:hours).distinct })
 
--- a/app/models/volunteer.rb
+++ b/app/models/volunteer.rb
@@ -50,6 +50,8 @@ class Volunteer < ApplicationRecord
   has_many :billing_expenses
   has_many :reminders, dependent: :destroy
 
+  has_and_belongs_to_many :group_offers
+
   has_attached_file :avatar, styles: { thumb: '100x100#' }
 
   validates :contact, presence: true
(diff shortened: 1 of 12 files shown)


app/models/volunteer.rb
```diff
@@ -71,6 +73,12 @@ class Volunteer < ApplicationRecord
     joins(:assignments).merge(Assignment.active_between(start_date, end_date)).distinct
   }
   scope :without_assignment, (-> { left_outer_joins(:assignments).where(assignments: { id: nil }) })
+  scope :without_active_assignment, (-> { joins(:assignments).merge(Assignment.ended) })
+  scope :not_responsible, (-> { where.not(id: GroupOffer.pluck(:responsible_id)) })
```
toupeira: This will run two queries, you can do it all in one by using `where.not(id: GroupOffer.select(:responsible_id))`

This generates a subquery which looks like `SELECT ... WHERE id IN (SELECT id ...)` (see Rails log or console)

app/models/volunteer.rb
```diff
@@ -71,6 +73,12 @@ class Volunteer < ApplicationRecord
...
+  scope :not_in_any_group_offer, lambda {
+    joins('LEFT JOIN group_offers_volunteers ON volunteers.id = group_offers_volunteers.volunteer_id')
```
toupeira: `left_joins(:group_offers)` should do the same

app/models/volunteer.rb
```diff
@@ -71,6 +73,12 @@ class Volunteer < ApplicationRecord
...
+      .where('group_offers_volunteers.volunteer_id IS NULL')
```
toupeira: `.where(group_offers: { volunteer_id: nil })` should do the same

app/views/group_offers/index.html.slim
```diff
@@ -39,6 +41,17 @@ h1= t_title
...
+              - if offer.responsible
+                - if offer.volunteer_responsible_state?
+                  = t("volunteer_responsible_state.#{offer.volunteer_responsible_state}")
+                  br
+                = link_to offer.responsible, volunteer_path(offer.responsible)
+            td
+              - if offer.volunteers.any?
+                - offer.volunteers.each do |volunteer|
+                  = link_to volunteer, volunteer_path(volunteer)
+                  br
```
toupeira: The `if offer.volunteers.any?` isn't really necessary here and might even trigger an additional `COUNT(*)` query, you can also just have the `.each` block and if there are no volunteers it won't output anything.

app/models/volunteer.rb
chrysalag: That looks for a `volunteer_id` at the `group_offers`, but there isn't any since the relationship is many to many. That's why I use the join table.

toupeira: Ah sorry I meant `.where(group_offers_volunteers: { volunteer_id: nil })` (I thought you can specify the association name but it's indeed the table name)

PR #239, opened 2017-09-27
Feature: group offer as pdf

chrysalag:
# [Story in Trello](https://trello.com/c/DwFYw9yM/5-3-als-freiwilligenverantwortliche-r-m%C3%B6chte-ich-eine-vereinbarung-f%C3%BCr-das-gruppenangebot-erstellen)

### Done for approval?

* ~[ ] Story under test~
* [x] Mobile works
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/group_offers_controller.rb
+++ b/app/controllers/group_offers_controller.rb
@@ -6,7 +6,15 @@ class GroupOffersController < ApplicationController
     @group_offers = policy_scope(GroupOffer)
   end
 
-  def show; end
+  def show
+    respond_to do |format|
+      format.html
+      format.pdf do
+        render pdf: "group_offer_#{@group_offer.id}", layout: 'pdf.pdf',
+        template: 'group_offers/show.html.slim', encoding: 'UTF-8'
+      end
+    end
+  end
 
   def new
     @group_offer = GroupOffer.new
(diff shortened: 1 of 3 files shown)


app/controllers/group_offers_controller.rb
```diff
@@ -6,7 +6,15 @@ def index
     @group_offers = policy_scope(GroupOffer)
   end
 
-  def show; end
+  def show
+    respond_to do |format|
+      format.html
+      format.pdf do
+        render pdf: "group_offer_#{@group_offer.id}", layout: 'pdf.pdf',
+        template: 'group_offers/show.html.slim', encoding: 'UTF-8'
```
casaper: You could also create a file `group_offers/show.pdf.slim` with `render 'group_offers/show.html.slim'`
It would render the template option obsolete

chrysalag: Yes, I saw that at the billing expenses pdf. For some reason, it doesn't work and produces a missing partial error.

casaper: Strange, but maybe the path was wrong, or something. But never mind. Woking is working. So passing template is also ok. I just thought in general, it would make checks like `- unles params[:format] == 'pdf'` obsolete, because the html could be different for both formats and the stuff that is the same could be rendered in. But never mind. It is not important.

chrysalag: Indeed. Didn't find the reason was not working the other way, so I did it like this. So approval? :)

PR #229, opened 2017-09-19
Features/reimplementing volunteer state

casaper:
# [Story in Trello]()

### Done for approval?

* [ ] Acceptance criteria are met
* [ ] Code quality checks are green
* [ ] Readme updated if needed
* [ ] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [ ] Translated
* [ ] Code is reviewed

### Done after the merge?

* [ ] Deployed to staging after the merge
* [ ] Tested manually on staging

PR #241, opened 2017-09-27
Features/ga index and filters

chrysalag:
# [Story in Trello](https://trello.com/c/tIlm0AhJ/8-5-die-verschiedenen-gruppenangebote-sollen-in-einer-%C3%BCbersicht-dargestellt-werden-k%C3%B6nnen-zusammen-mit-der-information-welche-fw-d)

### Done for approval?

* [x] Story under test
* ~[ ] Mobile works~
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/helpers/filter_dropdown_helper.rb
+++ b/app/helpers/filter_dropdown_helper.rb
@@ -99,7 +99,9 @@ module FilterDropdownHelper
 
   def translate_value(filter_attribute, t_scope)
     return t('all') if filter_attribute.blank?
-    unless filter_attribute.class == Department || filter_attribute.to_s.to_i != 0
+    unless filter_attribute.class == Department ||
+        filter_attribute.class == GroupOfferCategory ||
+        filter_attribute.to_s.to_i != 0
       return t(filter_attribute, t_scope)
     end
     filter_attribute.to_s
--- a/app/helpers/filter_dropdown_helper.rb
+++ b/app/helpers/filter_dropdown_helper.rb
@@ -99,11 +99,10 @@ module FilterDropdownHelper
 
   def translate_value(filter_attribute, t_scope)
     return t('all') if filter_attribute.blank?
-    unless filter_attribute.class == Department ||
-        filter_attribute.class == GroupOfferCategory ||
+    if [Department, GroupOfferCategory].include?(filter_attribute.class) ||
         filter_attribute.to_s.to_i != 0
-      return t(filter_attribute, t_scope)
+      return filter_attribute.to_s
     end
-    filter_attribute.to_s
+    t(filter_attribute, t_scope)
   end
 end
--- a/app/helpers/filter_dropdown_helper.rb
+++ b/app/helpers/filter_dropdown_helper.rb
@@ -99,6 +99,9 @@ module FilterDropdownHelper
 
   def translate_value(filter_attribute, t_scope)
     return t('all') if filter_attribute.blank?
-    t(filter_attribute, t_scope)
+    unless filter_attribute.class == Department || filter_attribute.to_s.to_i != 0
+      return t(filter_attribute, t_scope)
+    end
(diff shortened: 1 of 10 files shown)


app/helpers/filter_dropdown_helper.rb
```diff
@@ -99,6 +99,11 @@ def dropdown_toggler_options
 
   def translate_value(filter_attribute, t_scope)
     return t('all') if filter_attribute.blank?
-    t(filter_attribute, t_scope)
+    unless filter_attribute.class == Department ||
```
casaper:
```ruby
[Department, GroupOfferCategory].include? filter_attribute.class
```

casaper:
```ruby
if your_condition
  return filter_attribute.to_s
end
t(filter_attribute, t_scope)

File: app/views/group_offers/index.html.slim [hunk 2]
```diff
@@ -45,7 +52,7 @@ table.table.table-striped
                 br
         td= t("offer_state.#{offer.offer_state}") if offer.offer_state?
         td
-          - if offer.responsible
+          - if offer.responsible && !offer.external?
```
casaper: the other way arround because its cheaper: `!offer.external? && offer.responsible`

app/views/group_offers/show.html.slim
```diff
@@ -65,7 +65,7 @@ h1= @group_offer.title
             td= t_attr(:schedule_details)
             td= @group_offer.schedule_details
 
-- if @group_offer.responsible
+- if @group_offer.responsible && !@group_offer.external?
```
casaper: the other way arround because its cheaper: !offer.external? && offer.responsible

Or better make a method in the model

app/helpers/filter_dropdown_helper.rb
```diff
@@ -99,6 +99,11 @@ def dropdown_toggler_options
...
+        filter_attribute.class == GroupOfferCategory ||
+        filter_attribute.to_s.to_i != 0
+      return t(filter_attribute, t_scope)
+    end
+    filter_attribute.to_s
```
toupeira: Hmm this method is getting *really* hairy, @casaper is also adding some changes in https://github.com/panter/aoz-003/pull/229/files#diff-c96b9a7dfc4caea3be8d6e40b89c9ad2

It looks like one of the issues is that `filter_attribute` is sometimes a record (`list_filter_link`), and sometimes an attribute name (`toggler_text`)?

Please try to refactor this method together with @casaper, or if it's getting too confusing just add some tests to make sure it won't break :)

app/helpers/filter_dropdown_helper.rb
casaper: And why arent you checking for `filter_attribute.is_a?` ?
If its a symbol or a model. You can check for that. And in that case your condition will be shorter and it will be more generic

app/helpers/filter_dropdown_helper.rb
casaper: Yep, that would make sense. Or use it in the same manner as it was used before: by not passing records, just symbols and t_scope if needed.

app/helpers/filter_dropdown_helper.rb
chrysalag: I had tried that. But when filtering for departments (and categories I guess), it first checks if there is translation for the department name and then for the department id.

app/helpers/filter_dropdown_helper.rb
toupeira: I think it's already passing records in the current `develop` in `list_eq_filter_dropdown` (`item`) and `list_boolean_filter_dropdown` (`collection[index]`), which become `filter_attribute` in `list_filter_link`, which then gets passed to `translate_value`.

PR #244, opened 2017-09-28
add department manager to categories scope and add tests

nuritnt:
Upon request department managers were also added to create a group offer category

### DIFF
--- a/app/policies/group_offer_category_policy.rb
+++ b/app/policies/group_offer_category_policy.rb
@@ -3,16 +3,18 @@ class GroupOfferCategoryPolicy < ApplicationPolicy
     def resolve
       if superadmin?
         all
+      elsif department_manager?
+        all
       else
         none
       end
     end
   end
 
-  alias_method :index?,                 :superadmin?
-  alias_method :show?,                  :superadmin?
-  alias_method :new?,                   :superadmin?
-  alias_method :edit?,                  :superadmin?
-  alias_method :create?,                :superadmin?
-  alias_method :update?,                :superadmin?
+  alias_method :index?,                 :superadmin_or_department_manager?
+  alias_method :show?,                  :superadmin_or_department_manager?
+  alias_method :new?,                   :superadmin_or_department_manager?
+  alias_method :edit?,                  :superadmin_or_department_manager?
+  alias_method :create?,                :superadmin_or_department_manager?
+  alias_method :update?,                :superadmin_or_department_manager?
 end
--- a/app/policies/group_offer_category_policy.rb
+++ b/app/policies/group_offer_category_policy.rb
@@ -1,16 +1,4 @@
 class GroupOfferCategoryPolicy < ApplicationPolicy
-  class Scope < ApplicationScope
-    def resolve
-      if superadmin?
-        all
-      elsif department_manager?
-        all
-      else
-        none
-      end
-    end
-  end
-
   alias_method :index?,                 :superadmin_or_department_manager?
   alias_method :show?,                  :superadmin_or_department_manager?
   alias_method :new?,                   :superadmin_or_department_manager?
--- a/test/policies/group_offer_categories_policy_test.rb
+++ b/test/policies/group_offer_categories_policy_test.rb
@@ -6,8 +6,8 @@ class GroupOfferCategoryPolicyTest < PolicyAssertions::Test
       'update?')
   end
 
-  test "department manager can't create/access categories" do
-    refute_permit(create(:department_manager), GroupOfferCategory, 'new?', 'create?', 'index?', 'show?',
-      'edit?', 'update?')
+  test 'department can use all category actions' do
+    assert_permit(create(:department_manager), GroupOfferCategory, 'new?', 'create?', 'index?',
+      'show?', 'edit?', 'update?')
   end
 end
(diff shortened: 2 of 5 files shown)


app/policies/group_offer_category_policy.rb
```diff
@@ -3,16 +3,18 @@ class Scope < ApplicationScope
     def resolve
       if superadmin?
         all
+      elsif department_manager?
```
nuritnt: does it make sense to add this resolve all like this since they both have the same policy or would it make sense to remove this resolve method and then adjust de controller?

chrysalag: I don't think we need resolve anymore.

app/views/group_offers/index.html.slim
```diff
@@ -3,7 +3,7 @@ h1= t_title
 nav.navbar.section-navigation
   ul.list-inline
     li= button_link t_title(:new), new_group_offer_path, 'success', dimension: 'sm'
-    - if policy(GroupOfferCategory).superadmin?
+    - if policy(GroupOfferCategory).superadmin_or_department_manager?
```
chrysalag: I think that is now solved from the GroupOffer policies

test/system/group_offer_categories_test.rb
```diff
@@ -3,11 +3,14 @@
...
     @user = create :user
-    login_as @user
+    @department = create :department
+    @department_manager = create :user, :with_departments
+    @group_offer_category = create :group_offer_category, category_name: 'Schwimmen'
   end
 
 
-  test 'create new group offer category and update it' do
+  test 'superadmin can create new group offer' do
```
chrysalag: Since these tests are pretty much the same in pairs, please check if you can iterate over the users and make the 4 tests 2. In that case printing a text at the beginning of each test with the logged in user would be helpful. Otherwise checking only for the superadmin should be enough since the ability to create or update a category is solved through policies.

app/policies/group_offer_category_policy.rb
toupeira: Yes I think we can probably remove the scope, only superadmins and department managers can access the `index` action so there's no need to restrict the scope further.

An exception would be if the group offer categories are listed in other controllers/views, but that doesn't seem to be the case here.

app/views/group_offers/index.html.slim
toupeira: Not sure, but in any case you should check for the action (`index?`) here rather than the role, otherwise you're duplicating the logic again.

In other words: you're rendering a link to the `index` action with `group_offer_categories_path`, so you should check if the user has permission for this action (`policy(GroupOfferCategory).index?`), instead of repeating the check for the user's role (which is already in `GroupOfferCategoryPolicy#index?`).

test/system/group_offer_categories_test.rb
toupeira: Instead of looping through the users inside the tests you can also define the tests dynamically *inside* the loop :)

```ruby
users = {
  'superadmin' => create(:user),
  'department manager' => create(:user, :with_departments),
}

users.each do |label, user|
  test "#{label} can update a category" do
    login_as user
    ...
  end

  test "#{label} can create new group offer" do
    login_as user
    ...
  end
end
```

This will give you the same 4 tests with correct error messages etc.

chrysalag: Nice! One small comment though for the implementation, :with_departments doesn't change the user role, so role: 'department_manager' should be added, as well.

nuritnt: @toupeira: chrysa and me tried to set the up the tests like this, but couldn't get it working. I decided to simply reduce the tests, since we check the users already in the model test.

test/policies/group_offer_categories_policy_test.rb
```diff
@@ -6,8 +6,8 @@ class GroupOfferCategoryPolicyTest < PolicyAssertions::Test
       'update?')
   end
 
-  test "department manager can't create/access categories" do
-    refute_permit(create(:department_manager), GroupOfferCategory, 'new?', 'create?', 'index?', 'show?',
-      'edit?', 'update?')
+  test 'department can use all category actions' do
```
chrysalag: NIT: department manager

test/system/group_offer_categories_test.rb
```diff
@@ -3,11 +3,11 @@
 class GroupOfferCategoriesTest < ApplicationSystemTestCase
   def setup
     @user = create :user
-    login_as @user
   end
 
 
-  test 'create new group offer category and update it' do
+  test 'superadmin can create new group offer' do
```
chrysalag: Since superadmin and department manager have the same rights, both tests could be for the superadmin.

app/views/group_offers/index.html.slim
```diff
@@ -3,7 +3,7 @@ h1= t_title
 nav.navbar.section-navigation
   ul.list-inline
     li= button_link t_title(:new), new_group_offer_path, 'success', dimension: 'sm'
-    - if policy(GroupOfferCategory).superadmin?
+    - if policy(GroupOfferCategory).index?
```
chrysalag: I am not sure if this is needed.

casaper: I agree, it isn't as the authorize in the controller will only let superadmin and department managers access the whole action

PR #246, opened 2017-09-28
group offer volunteers removal/edit

nuritnt:
# [Story in Trello](https://trello.com/c/4SFGVvWj/22-2-einzelne-freiwillige-k%C3%B6nnen-aus-einem-gruppenangebot-enfernt-werden)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Translated
* [ ] Code is reviewed

### Done after the merge?


nuritnt:
@toupeira @chrysalag @casaper 
this would be the idea how i would implement it, but would kinda need some support - I am not sure how to implement all by myself. 
- when you create a group offer, have a list of already chosen volunteers
- button if you want to edit the list of volunteers
- partial with index list of already chosen volunteers
- if you click on add more volunteers, the index with available volunters will be shown

has anybody time to support me or if anybody has a easier suggestion, i am really open for:)

chrysalag:
For the time being, I think that a first step could be to add the possibility to add/remove volunteers through the index page but not change the form. Reason is that the functionality exists, the only issue is the user friendliness. A nice way would be like it is now for languages with add and remove. Had tried that when adding one or more volunteers at the group offers with no success. We could try that together, if you like.

nuritnt:
what do you think about this?

config/locales/en.yml
```diff
@@ -564,6 +566,7 @@ en:
           c: 'C'
         state: *id-states
       group_offer:
+        add_volunteers: add volunteers
```
chrysalag: I think that is not needed.

config/locales/en.yml
```diff
@@ -261,6 +262,7 @@ en:
   add_journal_entry: Write a journal entry
   add_language: Add language
   add_relative: Add family member
+  add_volunteers: add volunteers
```
chrysalag: I think that is not needed.

config/locales/de.yml
```diff
@@ -734,6 +735,7 @@ de:
         salutation: *id-salutations
         state: *id-states
       group_offer:
+        add_volunteers: Freiwillige hinzufügen
```
chrysalag: Not needed. Only needed in the path (en/de).group_offers.form.add_volunteers

config/locales/de.yml
```diff
@@ -420,6 +420,7 @@ de:
   add_journal_entry: Journaleintrag verfassen
   add_language: Sprache hinzufügen
   add_relative: Verwandte hinzufügen
+  add_volunteers: Freiwillige hinzufügen
```
chrysalag: Not needed.

app/views/group_offers/_form.html.slim
```diff
@@ -72,6 +62,28 @@
...
+      = f.input :necessary_volunteers, collection: 1..10
+    - if @group_offer.volunteers.any?
+    .col-xs-12.col-md-4
+      =t_attr(:volunteers)
+      ul
+        - @group_offer.volunteers.each do |volunteer|
+          li= link_to volunteer, volunteer_path(volunteer)
+    .col-xs-12.col-md-4
+      button.btn.btn-primary(data-toggle='collapse' data-target='#volunteerBlock' type='button'  aria-expanded='false' aria-controls='volunteerBlock')
+          = 'edit Volunteers'
```
chrysalag: = t('.add_volunteers')

app/views/group_offers/_form.html.slim
```diff
@@ -26,16 +26,6 @@
...
-    .col-xs-12.col-md-8
-      - if policy(GroupOffer).supervisor_privileges?
-        = f.association :volunteers
-      - else
-        = f.association :volunteers,
-          collection: Volunteer.without_active_assignment.not_in_any_group_offer.not_responsible
-
-  .row
-    .col-xs-12.col-md-4
       - if policy(GroupOffer).supervisor_privileges?
```
chrysalag: Although that was not introduced in this PR, has to be removed. Department managers can also add volunteer responsible.

casaper: I think the opposite. Thats what a policy is for. 
This  policy:

```ruby
alias_method :supervisor_privileges?, :superadmin?
```

can be switched to:

```ruby
alias_method :supervisor_privileges?, :superadmin_or_department_manager?
```

So the policy can be adapted, and the views can stay the same.

toupeira:
Looks good code-wise 👍 Let me know if you still need help with the implementation, I probably have some time tomorrow or Thursday.

app/views/group_offers/_form.html.slim
```diff
@@ -26,17 +26,7 @@
...
-      - if policy(GroupOffer).supervisor_privileges?
-        = f.association :volunteers
-      - else
-        = f.association :volunteers,
-          collection: Volunteer.without_active_assignment.not_in_any_group_offer.not_responsible
-
-  .row
-    .col-xs-12.col-md-4
-      - if policy(GroupOffer).supervisor_privileges?
+      - if policy(GroupOffer).superadmin_or_department_manager?
```
chrysalag: What should be checked here is: 
superadmin can select from all volunteers, department manager from the ones available for assignment/group offer.

config/locales/en.yml
```diff
@@ -70,6 +70,8 @@ en:
         archived: Archived
         deactivate: Deactivate
       form:
+        add_volunteers: add volunteers
```
chrysalag: Since there are suggested changes, otherwise that is pure NIT picking, it could be `Add volunteers` and the next `Edit volunteers`

app/views/group_offers/_form.html.slim
```diff
@@ -72,6 +62,31 @@
...
+      = f.input :necessary_volunteers, collection: 1..10
+    - if @group_offer.volunteers.any?
+    .col-xs-12.col-md-4
+      =t_attr(:volunteers)
+      ul
+        - @group_offer.volunteers.each do |volunteer|
+          li= link_to volunteer, volunteer_path(volunteer)
+    .col-xs-12.col-md-4
+      button.btn.btn-primary(data-toggle='collapse' data-target='#volunteerBlock' type='button'  aria-expanded='false' aria-controls='volunteerBlock')
+        - if action_name ==  'new'
```
chrysalag: It could be also the case that when created volunteers weren't added and they are as part of an edit action. So what checked could be `if action_name == 'new' || @group_offer.empty?`

app/views/group_offers/_form.html.slim
```diff
@@ -72,6 +62,31 @@
...
+          = t('.add_volunteers')
+        - else
+          = t('.edit_volunteers')
+    .col-xs-12.col-md-4
+      #volunteerBlock.collapse
```
casaper: Not 100% sure, but I think this can be done and that would result one less nested div here.

```slim
#volunteerBlock.collapse.col-xs-12.col-md-4
```

app/views/group_offers/_form.html.slim
nuritnt: just if i get it correct, but the changes i did in the group offer policy were correct - or not?

app/views/group_offers/_form.html.slim
```diff
@@ -72,6 +62,31 @@
...
+        - if policy(GroupOffer).supervisor_privileges?
```
nuritnt: but can this policy check stay or need this also be adapted? @chrysalag

chrysalag: The policy needs to be adapted, supervisor & dep manager see different collections

app/views/group_offers/_form.html.slim
nuritnt: tried it, not really helpful, since if the chosen volunteer list is empty, it uses this row, else the row is shifted..

app/views/group_offers/_form.html.slim
chrysalag: The policy needs to be adapted, supervisor & dep manager see different collections

app/views/group_offers/_form.html.slim
```diff
@@ -26,19 +26,9 @@
...
-      - if policy(GroupOffer).supervisor_privileges?
-        = f.association :volunteers
-      - else
-        = f.association :volunteers,
-          collection: Volunteer.without_active_assignment.not_in_any_group_offer.not_responsible
-
-  .row
-    .col-xs-12.col-md-4
-      - if policy(GroupOffer).supervisor_privileges?
+      - if policy(GroupOffer).superadmin?
```
chrysalag: `.superadmin?` and `department_manager?` policies are not defined. To me there is no reason to change the `superadmin_priviledges?` policy. Also there is no reason to have to ifs executed when only one of the cases will be true each time.

app/views/group_offers/_form.html.slim
```diff
@@ -72,6 +62,30 @@
...
+        - @group_offer.volunteers.each do |volunteer|
+          li= link_to volunteer, volunteer_path(volunteer)
+    .col-xs-12.col-md-4
+      button.btn.btn-primary(data-toggle='collapse' data-target='#volunteerBlock' type='button'  aria-expanded='false' aria-controls='volunteerBlock')
+        - if action_name ==  'new' || @group_offer.volunteers.empty?
+          = t('.add_volunteers')
+        - else
+          = t('.edit_volunteers')
+    #volunteerBlock.collapse.col-xs-12.col-md-4
+      - if policy(GroupOffer).superadmin?
```
chrysalag: `.superadmin?` and `department_manager?` policies are not defined. To me there is no reason to change the `superadmin_priviledges?` policy. Also there is no reason to have to ifs executed when only one of the cases will be true each time.

PR #250, opened 2017-09-29
Implement restrictions for external volunteers

casaper:
# [Story in Trello](https://trello.com/c/X6kqnLvv/26-0-einschr%C3%A4nkungen-f%C3%BCr-externe-freiwillige)

### Done for approval?

* [x] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/volunteers_controller.rb
+++ b/app/controllers/volunteers_controller.rb
@@ -72,7 +72,12 @@ class VolunteersController < ApplicationController
   end
 
   def invite_volunteer_user
-    User.new_volunteer(@volunteer) && @volunteer.user.invite!
+    return false if @volunteer.external
+    new_user = User.new(
+      email: @volunteer.contact.primary_email, password: Devise.friendly_token,
+      role: 'volunteer', volunteer: @volunteer
+    )
+    new_user.save && new_user.invite!
   end
 
   def set_volunteer
--- a/app/models/volunteer.rb
+++ b/app/models/volunteer.rb
@@ -9,6 +9,9 @@ class Volunteer < ApplicationRecord
 
   before_validation :handle_external
   before_save :record_acceptance_changed
+  before_validation :handle_user_with_external_change,
+    if: :external_changed?,
+    unless: proc { user_id.nil? }
 
   SINGLE_ACCOMPANIMENTS = [:man, :woman, :family, :kid, :unaccompanied].freeze
   GROUP_ACCOMPANIMENTS = [:sport, :creative, :music, :culture, :training, :german_course,
@@ -52,7 +55,9 @@ class Volunteer < ApplicationRecord
     content_type: /\Aimage\/.*\z/
   }
 
-  validates :user, absence: true, if: :external?
+  validates :user, absence: true,
+    if: :external?,
+    unless: proc { |user| user.deleted? }
 
   default_scope { order(created_at: :desc) }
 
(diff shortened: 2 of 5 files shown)


app/controllers/volunteers_controller.rb
```diff
@@ -72,7 +72,12 @@ def handle_volunteer_update
   end
 
   def invite_volunteer_user
-    User.new_volunteer(@volunteer) && @volunteer.user.invite!
+    return false if @volunteer.external
+    new_user = User.new(
+      email: @volunteer.contact.primary_email, password: Devise.friendly_token,
+      role: 'volunteer', volunteer: @volunteer
```
toupeira: IMHO it's more readable to put each key on its own line 👮

app/models/volunteer.rb
```diff
@@ -176,4 +183,13 @@ def to_s
   def record_acceptance_changed
     self["#{acceptance_change_to_be_saved[1]}_at".to_sym] = Time.zone.now if will_save_change_to_acceptance?
   end
+
+  def external_changed?
+    will_save_change_to_external?
```
toupeira: Hmm looks like the `#{attribute}_changed?` methods are [deprecated/renamed in Rails 5.1](https://github.com/rails/rails/pull/25337#issuecomment-225166796), but the [documentation isn't updated yet](http://api.rubyonrails.org/classes/ActiveModel/Dirty.html).

Did you add this to avoid the deprecation warning? In the long-term it's probably better to just use the new names, even though they're long and ugly :(

app/models/volunteer.rb
```diff
@@ -176,4 +183,13 @@ def to_s
...
+  end
+
+  def handle_user_with_external_change
+    user.destroy if external?
+    user.restore if internal?
```
toupeira: You can also do `user.destroy if external? else user.restore`, or use a multi-line `if` / `else` ;-)

But I really wouldn't write it with two `if` statements because the two conditions exclude each other, but I won't realize that until I actually look at the definitions of `external?` and `internal?`

test/models/reminder_test.rb
```diff
@@ -36,4 +36,12 @@ def setup
     Reminder.conditionally_create_reminders
     assert_equal 10, Reminder.count
   end
+
+  test 'reminders can only be created with internal volunteers' do
+    assignment = create :assignment
+    assignment.volunteer.update(external: true)
+    assignment.reload
+    reminder = build :reminder, volunteer: assignment.volunteer
+    refute reminder.valid?
```
toupeira: You should also test for specific errors, something like `assert reminder.errors[:volunteer].present?`

app/models/volunteer.rb
toupeira: If you want to keep it you can also use `alias_method` as a shortcut

casaper: @chrysalag gave me the hint, I wasn't aware they're deprecated. But yea, the new naming isn't nice, I agree with that. But as this isn't a thing I'm going to look at all the time, I don't mind too much... 😉

app/controllers/volunteers_controller.rb
casaper: IMHO that would be cool, but RuboCop, the very cop will definiteley complain about method length if I did that.

PR #252, opened 2017-09-30
fix/client mail update

nuritnt:
# [Story in Trello](https://trello.com/c/3GJLklmX/44-bug-if-i-register-a-client-the-currentuser-mail-gets-overwritten)

ensuring that if a client mail is updated, the method `update_user_email` is not used, because a client is not a user.

### Done for approval?

* [x] Story under test
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -95,4 +95,15 @@ class UserTest < ActiveSupport::TestCase
     @user.reload
     assert_equal 'supervolunteer@example.com', @user.email
   end
+
+  test 'updating client mail doesnt change current user email' do
+    @superadmin = create :user, role: 'superadmin', email: 'superadmin@example.com'
+    @client = create :user, email: 'client@example.com'
+    login_as(@superadmin)
+    @client.profile.contact.update(primary_email: 'superclient@example.com')
+    @client.reload
+    assert_equal 'superclient@example.com', @client.profile.contact.primary_email
+    assert_not_equal 'superclient@example.com', @superadmin.email
+    assert_equal 'superadmin@example.com', @superadmin.email
+  end
 end
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -97,13 +97,12 @@ class UserTest < ActiveSupport::TestCase
   end
 
   test 'updating client mail doesnt change current user email' do
-    @superadmin = create :user, role: 'superadmin', email: 'superadmin@example.com'
-    @client = create :user, email: 'client@example.com'
-    login_as(@superadmin)
-    @client.profile.contact.update(primary_email: 'superclient@example.com')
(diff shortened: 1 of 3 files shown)


nuritnt:
any suggestions how to test this or rather how to write test for it?

toupeira:
> any suggestions how to test this or rather how to write test for it?

@nuritnt you could either test for the side-effects of `update_user_email`, or if it's too tricky you can use mocks/stubs to check if a callback is triggered or not. I'm not too familiar with Minitest but [here's an article](https://semaphoreci.com/community/tutorials/mocking-in-ruby-with-minitest), probably something for a Kata session :)

chrysalag:
You could write a model test to see if the bug you found can be now reproduced:
- Create a superadmin
- Create a client with the same email as the superadmin (although that should not be possible, anyway)
- Update the email client
- Check if the superadmin email is updated to the client's new email or if it is the same as it was.

test/models/user_test.rb
```diff
@@ -95,4 +95,15 @@ def setup
     @user.reload
     assert_equal 'supervolunteer@example.com', @user.email
   end
+
+  test 'updating client mail doesnt change current user email' do
+    @superadmin = create :user, role: 'superadmin', email: 'superadmin@example.com'
```
chrysalag: There is already `@superadmin` at the setup. You could have only added her email here. (Although by default the `create :user` creates a superadmin, I don't know why we need two.)

test/models/user_test.rb
```diff
@@ -95,4 +95,15 @@ def setup
...
+    @client = create :user, email: 'client@example.com'
+    login_as(@superadmin)
+    @client.profile.contact.update(primary_email: 'superclient@example.com')
+    @client.reload
```
chrysalag: You could also `@superadmin.reload`, just to be sure.

test/models/user_test.rb
```diff
@@ -95,4 +95,17 @@ def setup
...
+  test 'updating client mail doesnt change current user email' do
+    @superadmin.update(email: 'superadmin@example.com')
+    @superadmin.reload
+    @client = create :user, email: 'client@example.com'
+    login_as(@superadmin)
+    @client.profile.contact.update(primary_email: 'superclient@example.com')
+    @client.reload
+
+    assert_equal 'superclient@example.com', @client.profile.contact.primary_email
+    assert_not_equal 'superclient@example.com', @superadmin.email
```
toupeira: You can remove this line since you have an `assert_equal ..., @superadmin.email` below :)

toupeira: Actually you can remove `@superadmin` completely from this test, since you're already testing that case in the existing test above and only really need to verify the behaviour for clients here :)

PR #253, opened 2017-10-03
adapt policys tests for allowing department_manager to create assignment

casaper:
# [Story in Trello](https://trello.com/c/wr8lKTie/6-3-als-freiwilligenverantwortliche-r-m%C3%B6chte-ich-ein-tandem-inkl-vereinbarung-erstellen)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [ ] Seeds created
* [x] Translated
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -57,7 +57,7 @@ class AssignmentsController < ApplicationController
 
   def find_volunteer
     @client = Client.find(params[:id])
-    @q = Volunteer.seeking_clients.ransack(params[:q])
+    @q = policy_scope(Volunteer).seeking_clients.ransack(params[:q])
     @seeking_clients = @q.result
     authorize Assignment
   end
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -28,9 +28,9 @@ class AssignmentsController < ApplicationController
 
   def create
     @assignment = Assignment.new(assignment_params.merge(creator_id: current_user.id))
-    @assignment.client.state = Client::RESERVED
     authorize @assignment
     if @assignment.save
+      @assignment.client.update(state: Client::RESERVED)
       redirect_to assignments_url, make_notice
     else
       render :new
@@ -46,9 +46,7 @@ class AssignmentsController < ApplicationController
   end
 
   def destroy
-    client = @assignment.client
-    @assignment.destroy
-    client.update(state: Client::REGISTERED)
+    @assignment.destroy.client.update(state: Client::REGISTERED)
     redirect_to assignments_url, make_notice
   end
 
@@ -58,9 +56,8 @@ class AssignmentsController < ApplicationController
   end
 
   def find_client
-    @q = Client.need_accompanying.ransack(params[:q])
-    @need_accompanying = @q.result
-    @need_accompanying = @need_accompanying.paginate(page: params[:page])
+    @q = policy_scope(Client).need_accompanying.ransack(params[:q])
+    @need_accompanying = @q.result.paginate(page: params[:page])
   end
 
   private
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -39,11 +39,7 @@ class AssignmentsController < ApplicationController
 
   def update
     if @assignment.update(assignment_params)
-      if current_user.superadmin? || current_user.department_manager?
-        redirect_to assignments_url, make_notice
-      else
-        redirect_to @assignment.volunteer, make_notice
-      end
+      redirect_to(volunteer? ? @assignment.volunteer : assignments_url, make_notice)
     else
       render :edit
     end
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -3,8 +3,7 @@ class AssignmentsController < ApplicationController
 
   def index
     authorize Assignment
-    @assignments = policy_scope(Assignment)
-    @q = Assignment.ransack(params[:q])
+    @q = policy_scope(Assignment).ransack(params[:q])
     @assignments = @q.result.default_order.paginate(page: params[:page])
   end
 
@@ -38,7 +37,7 @@ class AssignmentsController < ApplicationController
 
   def update
     if @assignment.update(assignment_params)
-      if current_user.superadmin?
+      if current_user.superadmin? || current_user.department_manager?
         redirect_to assignments_url, make_notice
       else
         redirect_to @assignment.volunteer, make_notice
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -48,8 +48,7 @@ class AssignmentsController < ApplicationController
   def destroy
     client = @assignment.client
     @assignment.destroy
-    client.state = Client::REGISTERED
-    client.save
+    client.update(state: Client::REGISTERED)
     redirect_to assignments_url, make_notice
   end
 
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -1,5 +1,7 @@
 class AssignmentsController < ApplicationController
   before_action :set_assignment, only: [:show, :edit, :update, :destroy]
+  before_action :set_volunteer, only: [:find_client]
+  before_action :set_client, only: [:find_volunteer]
 
   def index
     authorize Assignment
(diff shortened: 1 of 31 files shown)


app/policies/application_policy.rb
```diff
@@ -79,6 +79,14 @@ def superadmin_or_volunteer_related?
     superadmin? || volunteer_related?
   end
 
+  def admin_or_department_manager_or_assignment_related?
+    superadmin_or_department_manager? || volunteer_related?
+  end
+
+  def admin_or_department_manager_or_volunteer_related?
```
chrysalag: The naming of this function and of the above one are a bit confusing. There is `volunteer_related` in the title here but `volunteer_related?` is called in the above function. Also (I don't know if RuboCop is a reason for that) but since we only have superadmins and not admins, maybe the naming should be `supeadmin_or...`

app/policies/assignment_policy.rb
```diff
@@ -1,18 +1,21 @@
 class AssignmentPolicy < ApplicationPolicy
   class Scope < ApplicationScope
     def resolve
-      all if superadmin?
+      return all if superadmin?
+      scope.where(creator_id: user.id) if department_manager?
```
chrysalag: I think (based on reviews of other PRs) that there should be a last case where it returns `none`. That concerns also the other policy files touched in this PR.

app/policies/client_policy.rb
```diff
@@ -2,19 +2,23 @@ class ClientPolicy < ApplicationPolicy
   class Scope < ApplicationScope
     def resolve
       return all if superadmin?
-      resolve_owner if social_worker?
```
chrysalag: If I got it right yesterday at the standup, we keep the social workers with their current right.

app/policies/volunteer_policy.rb
```diff
@@ -1,15 +1,29 @@
...
-  alias_method :new?,             :superadmin?
-  alias_method :create?,          :superadmin?
+  class Scope < ApplicationScope
+    def resolve
+      return all if superadmin?
+      seeking_clients if department_manager?
+    end
+
+    def seeking_clients
+      return scope.seeking_clients_will_take_more if superadmin?
```
chrysalag: If this def is called `if department_manager?` how is this line going to be called?

config/locales/de.yml
```diff
@@ -152,7 +152,8 @@ de:
       create: &id-volunteers-create
         <<: *id-availability
         acceptance: *id-volunteer-acceptance
-        availability: Wann sind Sie für einen Freiwilligeneinsatz verfügbar?
+        availability: Verfügbarkeit?
+        availability_question: Wann sind Sie für einen Freiwilligeneinsatz verfügbar?
```
chrysalag: Since this translation was introduces here, should there be a call for this in the form in this PR?

app/controllers/volunteers_controller.rb
```diff
@@ -56,7 +56,7 @@ def destroy
 
   def seeking_clients
     authorize Volunteer
-    @q = Volunteer.seeking_clients.ransack(params[:q])
+    @q = specific_policy_scope(:seeking_clients).ransack(params[:q])
```
toupeira: Is this the same as `policy_scope(Volunteer).seeking_clients`? Because that would be a lot easier to follow ;-)

app/controllers/assignments_controller.rb
```diff
@@ -38,7 +37,7 @@ def create
 
   def update
     if @assignment.update(assignment_params)
-      if current_user.superadmin?
+      if current_user.superadmin? || current_user.department_manager?
```
toupeira: Maybe you could also do `authorize @assignment` before the `if` and do the redirect in a `rescue NotAuthorizedError`? Not sure if it's really better but I think it would be good to keep the rules in the policy.

app/models/assignment.rb
```diff
@@ -31,16 +31,16 @@ class Assignment < ApplicationRecord
   scope :no_end, (-> { where(period_end: nil) })
   scope :has_end, (-> { where.not(period_end: nil) })
 
-  scope :ended, (-> { where('period_end < ?', Time.zone.now) })
+  scope :ended, (-> { where('period_end < ?', Time.zone.now.to_date) })
```
toupeira: Wanted to say `Date.today` but that doesn't use the time zone from Rails. But there's also `Time.zone.today` which returns a `Date`

app/models/volunteer.rb
```diff
@@ -102,11 +106,18 @@ class Volunteer < ApplicationRecord
...
       .merge(Assignment.inactive)
       .where.not(assignments: { volunteer_id: with_active_assignments.ids })
       .or(loj_without_assignments)
+  }
+  scope :seeking_clients_will_take_more, lambda {
+    accepted_joined
+      .merge(Assignment.inactive)
+      .where.not(assignments: { volunteer_id: with_active_assignments.ids })
+      .or(loj_without_assignments)
       .or(loj_active_take_more)
```
toupeira: Not sure if the combination of not/or/or will really do what you want, maybe it does though ;-)

app/policies/assignment_policy.rb
toupeira: Yeah I think that would make sense :)

app/policies/volunteer_policy.rb
toupeira: If I understand correctly this is possible through `VolunteersController#seeking_clients`

app/models/assignment.rb
casaper: Ah, thats a great info. Rubocop was always bashing me if I did `Date.today`, so I used `Time.zone.now.to_date`, but probably RuboCop might like `Time.zone.today` better ;) Thanks.

app/models/volunteer.rb
casaper: I made extensive tests that made me believe it works. But I'm obviously open to better solutions. Incompatible queries made this scopes to be quite a journey. 😉

config/locales/de.yml
casaper: The translation was missing, so it hat to be put in place.

chrysalag: For some reason I didn't think of that case :P

app/policies/application_policy.rb
```diff
@@ -23,6 +23,14 @@ def superadmin_or_department_manager?
     superadmin? || department_manager?
   end
 
+  def department_manager_or_social_worker?
+    department_manager? || social_worker?
+  end
+
+  def volunteer_managing_user?
```
chrysalag: Not important probably, maybe user_managing_volunteer?

app/policies/application_policy.rb
```diff
@@ -79,6 +87,10 @@ def superadmin_or_volunteer_related?
     superadmin? || volunteer_related?
   end
 
+  def admin_or_department_manager_or_assignment_related?
```
chrysalag: `...._or_volunteer_related?`

app/policies/client_policy.rb
```diff
@@ -2,19 +2,25 @@ class ClientPolicy < ApplicationPolicy
...
-  alias_method :create?,            :superadmin_or_social_worker?
+  # controller action policies
+  alias_method :index?,  :volunteer_managing_user?
+  alias_method :new?,    :volunteer_managing_user?
+  alias_method :create?, :volunteer_managing_user?
 
-  alias_method :show?,              :superadmin_or_social_workers_record?
-  alias_method :edit?,              :superadmin_or_social_workers_record?
-  alias_method :update?,            :superadmin_or_social_workers_record?
+  alias_method :show?,   :superadmin_or_social_workers_record?
```
chrysalag: Not sure, but since we want department managers to create assignments, shouldn't they be able to see their details in order to decide for an assignment?

config/locales/de.yml
```diff
@@ -352,7 +353,7 @@ de:
         role: Rolle
       volunteer: &id-volunteer_attributes
         <<: [*id-availability, *id-generic_keys, *id-volunteer-acceptance]
-        acceptance: Affirmation
+        acceptance: *id-volunteer-acceptance
```
chrysalag: If `acceptance: Affirmation` gets removed then we'll have Acceptance again in the filter and the form. I also thought that this line was covered by this one: `<<: [*id-availability, *id-generic_keys, *id-volunteer-acceptance]`.

test/system/socialworker_test.rb
```diff
@@ -16,27 +16,12 @@ def setup
     assert_not page.has_field? 'Role'
   end
 
-  test 'has a navbar link to clients page' do
```
chrysalag: maybe add it back since the policy is restored?

test/system/socialworker_test.rb
```diff
@@ -16,27 +16,12 @@ def setup
...
-    assert page.has_link? 'Clients'
-  end
-
   test 'does not have navbar link to users and volunteers page' do
     visit user_path(@socialworker.id)
     assert_not page.has_link? 'Users'
     assert_not page.has_link? 'Volunteers'
   end
 
-  test 'can see his clients' do
```
chrysalag: maybe add it back since the policy is restored?

config/locales/de.yml
casaper: True

PR #259, opened 2017-10-06
Volunteer can give feedback to group_assignments and assignments

casaper:
# [Story in Trello](https://trello.com/c/s5cEa1Kq/15-3-als-frewillige-r-m%C3%B6chte-ich-feedback-zu-meinem-gruppeneinsatz-geben)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Translated
* [x] Code is reviewed

### Done after the merge?


app/views/feedbacks/show.html.slim
```diff
@@ -0,0 +1,29 @@
...
+          td= t_model(GroupOffer)
+          td= link_to @feedback.feedbackable.to_label, group_offer_path(@feedback.feedbackable)
+      - [:goals, :achievements, :future, :comments].each do |field|
+        tr
+          td= t_attr(field)
+          td= @feedback[field]
+      tr
+        td= t_attr(:conversation)
+        td= t(@feedback.conversation)
+
```
nuritnt: when logged in as a `superadmin`:
- creating a new feedback throws an error on submit
- should a superadmin be able to create even one? if not, a superadmin should see `new feedback` action button
- volunteers profile -> clicks on group offer list on `feedback index` shows an additional other volunteer which is not connected to the group offer

logged in as a `volunteer` the issues above mentioned are working properly

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    if params[:group_offer_id]
+      @feedbacks = @feedbacks.where(feedbackable_id: params[:group_offer_id])
+      @volunteer = current_user.volunteer || @feedbacks.first.volunteer
+    elsif params[:assignment_id]
+      @feedbacks = @feedbacks.where(feedbackable_id: params[:assignment_id])
+      @volunteer = Assignment.find(params[:assignment_id]).volunteer
+    elsif params[:volunteer_id]
+      @feedbacks = @feedbacks.where(volunteer_id: params[:volunteer_id])
+      @volunteer = Volunteer.find(params[:volunteer_id])
+    end
```
toupeira: Hmm it looks like this could be simplified to `params[:feedbackable]` and something like `@feedbackable = @feedbacks.first.feedbackable`

Not sure how important the `current_user.volunteer` case is though.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    authorize @feedback
+  end
+
+  def edit; end
+
+  def create
+    @feedback = Feedback.new(feedback_params.merge(author_id: current_user.id))
+    authorize @feedback
+    if @feedback.save
+      redirect_to url_for([@feedback.volunteer, @feedback]), make_notice
```
toupeira: Not sure if `@feedback.volunteer` is correct for the cases where the feedbackable model isn't a volunteer. Maybe use `@feedback.feedbackable` as well?

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    else
+      render :new
+    end
+  end
+
+  def update
+    if @feedback.update(feedback_params)
+      redirect_to url_for([@feedback.volunteer, @feedback]), make_notice
```
toupeira: Same here, volunteer vs. feedbackable

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  private
+
+  def set_feedbackable
+    @feedbackable = Assignment.find(params[:assignment_id]) if params[:assignment_id]
+    @feedbackable = GroupOffer.find(params[:group_offer_id]) if params[:group_offer_id]
+    return @volunteer = @feedbackable.volunteer if @feedbackable.class == Assignment
+    @volunteer = current_user.volunteer if current_user.volunteer?
+    @volunteer = Volunteer.find(params[:volunteer]) if params[:volunteer]
```
toupeira: Same concern here, but maybe I'm just missing something :)

app/models/billing_expense.rb
```diff
@@ -15,14 +15,9 @@ class BillingExpense < ApplicationRecord
...
-      self.amount = 100
-    elsif hour_count >= 1
-      self.amount = 50
-    else
-      self.amount = 0
-    end
+    return self.amount = 150 if hour_count > 50
+    return self.amount = 100 if hour_count > 25
+    return self.amount = 50 if hour_count >= 1
+    self.amount = 0
```
toupeira: Just repeating that I find these a *lot* harder to parse than a good old multi-line if statement :)

app/models/volunteer.rb
```diff
@@ -47,6 +48,11 @@ class Volunteer < ApplicationRecord
   has_many :journals, as: :journalable, dependent: :destroy
   accepts_nested_attributes_for :journals, allow_destroy: true
 
+  has_many :assignments, dependent: :destroy
+  has_many :clients, through: :assignments
+
+
+
```
toupeira: Remove empty lines 👮

app/views/feedbacks/index.html.slim
```diff
@@ -0,0 +1,34 @@
...
+
+  tbody
+    - @feedbacks.each do |record|
+      tr
+        td= link_to record.volunteer.contact.full_name, volunteer_path(record.volunteer)
+        td
+          - if record.assignment?
+            = link_to record.feedbackable.to_label, client_path(record.feedbackable.client)
+          - else
+            = link_to record.feedbackable.to_label, group_offer_path(record.feedbackable)
```
toupeira: Note that there's also `polymorphic_path`, maybe you can use it here or in other places. See http://thelazylog.com/polymorphic-routes-in-rails/

test/models/feedback_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+
+class FeedbackTest < ActiveSupport::TestCase
+  test 'feedback is valid' do
+    feedback = build :feedback
+    assert feedback.valid?
+  end
```
toupeira: Already covered by `test/models/factory_test.rb` :)

app/models/billing_expense.rb
casaper: Go and slap RuboCop for belling on it 😉

app/controllers/feedbacks_controller.rb
casaper: Its because if a superadmin wants to leave a feedback on a GroupOffer, there is no volunteer available. Thats why I passed it thru a get param in the new Feedback link. To avoid overlappings with :volunteer_id, i chose :volunteer
I find it strange that a superadmin can and should leave feedback, but AFIR superadmins are supposed to be able

app/controllers/feedbacks_controller.rb
casaper: so you consider `[@feedbackable, @feedback]`?

app/controllers/feedbacks_controller.rb
casaper: feedbackable are only `:group_offers` and `:assignments`. But Feedbacks have a relation to `:volunteer`.

app/views/feedbacks/index.html.slim
casaper: yep, the paths are all quite a bit of a lacky hacky thingie here.

app/models/volunteer.rb
casaper: 👍

app/controllers/feedbacks_controller.rb
toupeira: Ah ok, in that case ignore my comment :)

app/controllers/feedbacks_controller.rb
toupeira: Can be ignored as well I guess, since volunteer is not a feedbackable

config/locales/en.yml
```diff
@@ -209,12 +212,13 @@ en:
...
         period_end: End date
         period_start: Start date
-      assignment_journal:
+      feedback:
         achievements: What could have been achieved during the last months?
         conversation: I would like to discuss with my supervisor.
         future: Should the assignment continue? If yes, with which goals?
         goals: Which were the most important goals during the last months?
         last: Last assignment journal entry
+        feedbackable_id_and_type: Activities
```
toupeira: Do we need `feedback_activity` here as well? (added in `de`)

db/migrate/20171006133132_rename_assignment_journal_to_feedback.rb
```diff
@@ -0,0 +1,9 @@
+class RenameAssignmentJournalToFeedback < ActiveRecord::Migration[5.1]
+  def up
+    rename_table :assignment_journals, :feedbacks
+  end
+
+  def down
+    rename_table :feedbacks, :assignment_journals
+  end
```
toupeira: You don't really need the separate `up` / `down` methods, you can just define `change` and Rails knows how to reverse `rename_table` (see https://github.com/rails/rails/blob/master/activerecord/lib/active_record/migration/command_recorder.rb for a list of supported methods)

db/migrate/20171006140658_modify_feedbacks_to_be_polymorphic.rb
```diff
@@ -0,0 +1,17 @@
+class ModifyFeedbacksToBePolymorphic < ActiveRecord::Migration[5.1]
+  def up
+    add_reference :feedbacks, :feedbackable, polymorphic: true, index: true
+    Feedback.with_deleted.each do |feedback|
+      feedback.update(feedbackable_id: feedback.assignment_id, feedbackable_type: 'Assignment')
```
toupeira: `feedback.update(feedbackable: feedback.assignment)` should do the same (just FYI, you can leave it as it is)

app/controllers/feedbacks_controller.rb
toupeira: Are the params here used for filtering in the UI or something?

app/views/feedbacks/_main_fields.html.slim
```diff
@@ -0,0 +1,12 @@
+
+= f.hidden_field :volunteer_id, value: @feedback.volunteer.id
+= f.hidden_field :author_id, value: @feedback.author_id || current_user.id
```
toupeira: - the `value:` shouldn't really be necessary since it's already taken from `@feedback` by default
- you're already assigning `author` in the controller so `current_user` shouldn't be necessary here
- is it actually necessary to send `author_id` in the form, i.e. can a user create feedback for another user, or change the author of a feedback? If not, you could/should remove `author_id` from the form and `feedback_params` and only set the author in the `create` action

app/views/feedbacks/index.html.slim
```diff
@@ -0,0 +1,42 @@
...
+        td= record.goals
+        td= record.achievements
+        td= record.future
+        td= record.comments
+        td= t(record.conversation)
+        - unless current_user.volunteer?
+          td
+            - if record.author.id == record.volunteer.user.id
```
toupeira: You can also compare ActiveRecord objects directly (`record.author == record.volunteer.user`), by default it uses the `id` for comparison.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  private
+
+  def set_feedbackable
+    return @feedbackable = Assignment.find(params[:assignment_id]) if params[:assignment_id]
+    return @feedbackable = GroupOffer.find(params[:group_offer_id]) if params[:group_offer_id]
+    if params[:feedbackable_type] == 'GroupOffer'
+      @feedbackable = GroupOffer.find(params[:feedbackable_id])
+    end
```
toupeira: This method still seems a bit convoluted but I don't understand the feature well enough to say how it could be simplified :)

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def set_volunteer
+    @volunteer = current_user.volunteer if current_user.volunteer?
+    @volunteer ||= Volunteer.find(params[:volunteer_id]) if params[:volunteer_id]
+    @volunteer ||= Volunteer.find(params[:volunteer]) if params[:volunteer]
+    @volunteer ||= @feedbackable.volunteer if @feedbackable.class == Assignment
```
toupeira: This method still seems a bit convoluted but I don't understand the feature well enough to say how it could be simplified :)

app/controllers/feedbacks_controller.rb
casaper: I'm not yet sure how. The fact that superadmins can also add feedback complicates the whole thing.
Otherwise set_feedbackable and set_volunteer could be half as big.

app/models/billing_expense.rb
casaper: AFAIR it said I should use guard clauses. But well, I don't really care.

app/controllers/feedbacks_controller.rb
casaper: there is three different nested routes

```ruby
  resources :volunteers do
    resources :feedbacks
  end
  resources :assignments do
    resources :feedbacks
  end
  resources :group_offers do
    resources :feedbacks
  end
```

app/views/feedbacks/_main_fields.html.slim
casaper: > the value: shouldn't really be necessary since it's already taken from @feedback by default
you're already assigning author in the controller so current_user shouldn't be necessary here

Seems to make quite some sense 👍 

> is it actually necessary to send author_id in the form, i.e. can a user create feedback for another user, or change the author of a feedback? If not, you could/should remove author_id from the form and feedback_params and only set the author in the create action

the superadmins can create feedback on a volunteers GroupOffer or Assignment. Thats why there is a author field on Feedback at all, because if only the volunteer.user could create feedback on its own stuff, the field would render obsolete.

config/locales/en.yml
casaper: AFAIR I've just given this a different name so it won't be misleading. I chose some name in de.yml, but I' don't remember what exactly right now.

app/controllers/feedbacks_controller.rb
casaper: The problem is if superadmin clicks on `new_group_offer_feedback_path(group_offer)` I found no propper way to grab the `@volunteer` with. The only, admitingly not so beautiful solutiotion, that came in to mind, was passing it as a get parameter in that case.

casaper: In case of `Assignment` there is no problem, because `Assignment` `has_one` `Volunteer`, but `GroupOffer` `has_many`, so no way to get it from there in that case

db/migrate/20171006133132_rename_assignment_journal_to_feedback.rb
casaper: Ah, I guess I did that out of the intend to be a good boy, after RuboCop told me to do up and down in `20171006140658_modify_feedbacks_to_be_polymorphic.rb` 😉 
I'll change that.

app/controllers/feedbacks_controller.rb
casaper: Why did I not realize earlyer, that I could have just double nested the thing, to have all the information needed available 🤦‍♂️

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,69 @@
+class FeedbacksController < ApplicationController
+  before_action :set_feedback, only: [:show, :edit, :update, :destroy]
+  before_action :set_feedbackable
+  before_action :set_volunteer
+
+  def index
+    authorize Feedback
+    @feedbacks = policy_scope(Feedback)
+    @feedbacks = @feedbacks.where(feedbackable: @feedbackable)
```
chrysalag: Doesn't make a difference, but this could be in one line I guess.
`@feedbacks = policy_scope(Feedback).where(feedbackable: @feedbackable)`

app/models/feedback.rb
```diff
@@ -0,0 +1,30 @@
+class Feedback < ApplicationRecord
+  attr_reader :feedbackable_id_and_type
+
+  belongs_to :volunteer
+  belongs_to :author, class_name: 'User'
+  belongs_to :feedbackable, polymorphic: true, optional: true
+
+  validates :comments, presence: true
+
+  scope :on_group_offer, lambda { |group_offer_id|
```
chrysalag: Are these scopes used somewhere?

test/system/feedbacks_test.rb
```diff
@@ -0,0 +1,189 @@
...
+  test 'volunteer can see assignment feedbacks index' do
+    setup_feedbacks
+    login_as @user_volunteer
+    visit root_url
+    click_link 'volunteer@example.com'
+    click_link 'Show profile'
+    within '.assignments-table' do
+      click_link 'Feedback index'
+    end
+    refute page.has_text? 'author superadmin'
```
chrysalag: Maybe here you could also check if a text from the @user_volunteer exists, so as to check that it can indeed visit that page.

test/system/feedbacks_test.rb
```diff
@@ -0,0 +1,189 @@
...
+  test 'volunteer can see group offer feedbacks index' do
+    setup_feedbacks
+    login_as @user_volunteer
+    visit root_url
+    click_link 'volunteer@example.com'
+    click_link 'Show profile'
+    within '.group-assignments-table' do
+      click_link 'Feedback index'
+    end
+    refute page.has_text? 'author superadmin'
```
chrysalag: Same here

test/system/feedbacks_test.rb
```diff
@@ -0,0 +1,189 @@
...
+  test 'assignment feedback index contains only the feedbacks of one assignment' do
+    assignment2 = create :assignment, volunteer: @volunteer
+    create :feedback, volunteer: @volunteer, feedbackable: assignment2,
+      author: @superadmin, comments: 'assignment_number_2'
+    volunteer_feedback = create :feedback, volunteer: @volunteer, feedbackable: @assignment,
+      author: @user_volunteer, comments: 'assignment_number_1'
+    login_as @user_volunteer
+    visit polymorphic_path([@volunteer, @assignment, volunteer_feedback])
+    click_link 'Back'
+    refute page.has_text? 'assignment_number_2'
```
chrysalag: same here

app/controllers/feedbacks_controller.rb
casaper: true. Ill chaange it.

PR #261, opened 2017-10-06
Features/billing expenses and certificate for group offers

nuritnt:
# [Story in Trello](#)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/certificates_controller.rb
+++ b/app/controllers/certificates_controller.rb
@@ -58,7 +58,8 @@ class CertificatesController < ApplicationController
   def certificate_params
     params.require(:certificate).permit(
       :duration, :duration_end, :duration_start, :hours, :minutes, :text_body, :institution,
-      :function, volunteer_contact: [:name, :street, :city], assignment_kinds: Assignment::KINDS
+      :function, :group_offer, volunteer_contact: [:name, :street, :city],
+      assignment_kinds: Assignment::KINDS
     )
   end
 end
--- a/app/controllers/certificates_controller.rb
+++ b/app/controllers/certificates_controller.rb
@@ -1,5 +1,5 @@
 class CertificatesController < ApplicationController
-  before_action :set_certificate, except: [:show, :index, :new, :create]
+  before_action :set_certificate, only: [:show, :edit, :update, :destroy]
   before_action :set_volunteer
 
   def index
@@ -19,7 +19,7 @@ class CertificatesController < ApplicationController
   end
 
   def new
-    @certificate = Certificate.new(volunteer_id: @volunteer.id, user_id: current_user.id)
+    @certificate = Certificate.new(volunteer: @volunteer, user_id: current_user.id)
     @certificate.build_values
     authorize @certificate
   end
@@ -27,8 +27,9 @@ class CertificatesController < ApplicationController
   def edit; end
 
   def create
-    @certificate = Certificate.new(certificate_params.merge(volunteer_id: @volunteer.id,
+    @certificate = Certificate.new(certificate_params.except(:assignment_kinds).merge(volunteer: @volunteer,
       user_id: current_user.id))
+    @certificate.assignment_kinds = certificate_params.to_unsafe_h[:assignment_kinds]
     authorize @certificate
     if @certificate.save
       redirect_to volunteer_certificate_path(@volunteer, @certificate)
@@ -38,10 +39,10 @@ class CertificatesController < ApplicationController
   end
 
   def update
-    if @certificate.update(certificate_params)
+    if @certificate.update(certificate_params.except(:assignment_kinds)) && @certificate.update(assignment_kinds: certificate_params.to_unsafe_h[:assignment_kinds])
       redirect_to volunteer_certificate_path(@volunteer, @certificate)
     else
-      render :new
+      render :edit
     end
   end
 
@@ -54,18 +55,19 @@ class CertificatesController < ApplicationController
 
   def set_certificate
     @certificate = Certificate.find(params[:id])
+    @volunteer = @certificate.volunteer
     authorize @certificate
   end
 
   def set_volunteer
-    @volunteer = Volunteer.find(params[:volunteer_id]) if params[:volunteer_id]
+    @volunteer = Volunteer.find(params[:volunteer_id])
   end
 
   def certificate_params
     params.require(:certificate).permit(
       :duration, :duration_end, :duration_start, :hours, :minutes, :text_body, :institution,
-      :function, :group_offer, :volunteer_id, volunteer_contact: [:name, :street, :city],
-      assignment_kinds: Assignment::KINDS
+      :function, :volunteer_id, volunteer_contact: [:name, :street, :city],
+      assignment_kinds: [:group_offer, :assignment]
     )
   end
 end
--- a/app/controllers/certificates_controller.rb
+++ b/app/controllers/certificates_controller.rb
@@ -1,6 +1,6 @@
 class CertificatesController < ApplicationController
-  before_action :set_certificate, except: [:index, :new]
-  before_action :set_volunteer, only: [:index, :new]
+  before_action :set_certificate, except: [:show, :index, :new, :create]
+  before_action :set_volunteer
 
   def index
     authorize Certificate
@@ -54,18 +54,17 @@ class CertificatesController < ApplicationController
 
   def set_certificate
     @certificate = Certificate.find(params[:id])
-    @volunteer = @certificate.volunteer
     authorize @certificate
   end
 
   def set_volunteer
-    @volunteer = Volunteer.find(params[:volunteer_id])
+    @volunteer = Volunteer.find(params[:volunteer_id]) if params[:volunteer_id]
   end
 
   def certificate_params
     params.require(:certificate).permit(
       :duration, :duration_end, :duration_start, :hours, :minutes, :text_body, :institution,
-      :function, :group_offer, volunteer_contact: [:name, :street, :city],
+      :function, :group_offer, :volunteer_id, volunteer_contact: [:name, :street, :city],
       assignment_kinds: Assignment::KINDS
     )
   end
--- a/app/controllers/certificates_controller.rb
+++ b/app/controllers/certificates_controller.rb
@@ -20,18 +20,25 @@ class CertificatesController < ApplicationController
 
   def new
     @certificate = Certificate.new(volunteer_id: @volunteer.id, user_id: current_user.id)
+    @certificate.build_values
     authorize @certificate
   end
 
   def edit; end
 
-  def update
-    @certificate.update(certificate_params.merge(volunteer_id: @volunteer.id,
+  def create
+    @certificate = Certificate.new(certificate_params.merge(volunteer_id: @volunteer.id,
       user_id: current_user.id))
-    @certificate.update(volunteer_contact: certificate_params['volunteer_contact'],
-      assignment_kinds: certificate_params['assignment_kinds'])
     authorize @certificate
     if @certificate.save
+      redirect_to volunteer_certificate_path(@volunteer, @certificate)
+    else
+      render :new
+    end
+  end
+
+  def update
+    if @certificate.update(certificate_params)
       redirect_to volunteer_certificate_path(@volunteer, @certificate)
     else
       render :new
--- a/app/models/certificate.rb
+++ b/app/models/certificate.rb
@@ -21,14 +21,6 @@ class Certificate < ApplicationRecord
     }
   end
 
-  def convert_volunteer_contact
-    {
-      name: volunteer.contact.full_name,
-      street: volunteer.contact.full_street,
-      city: volunteer.contact.full_city
-    }
-  end
-
   def assignment_kinds=(value)
     super(value.to_h.map do |key, bool|
       bool = bool.to_i == 1 if bool.is_a? String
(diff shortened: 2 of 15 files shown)


app/models/certificate.rb
```diff
@@ -21,6 +21,14 @@ def convert_volunteer_contact
     }
   end
 
+  def convert_volunteer_contact
+    {
+      name: volunteer.contact.full_name,
+      street: volunteer.contact.full_street,
+      city: volunteer.contact.full_city
+    }
+  end
```
toupeira: This method already exists above? :)

app/views/certificates/_form.html.slim
```diff
@@ -17,6 +17,7 @@
       .row
         - Assignment::KINDS.each do |kind|
           .col-xs-2= fk.input kind, as: :boolean, input_html: { checked: @certificate.assignment_kinds[kind.to_s] }
+    = f.input :group_offer, as: :boolean, input_html: { checked: @certificate.group_offer }
```
toupeira: Hmm do you really need the `input_html`? It should already use the current value automatically.

app/views/certificates/show.html.slim
```diff
@@ -34,9 +34,11 @@
           span.kind-list
             input*{ type: 'checkbox', checked: value } onclick='return false;'
             = "#{t("assignment_kinds.#{kind}")}"
+        input*{ type: 'checkbox', checked: "#{@certificate.group_offer}" } onclick='return false;'
```
toupeira: I think you can simply do `input type="checkbox" checked="#{@certificate.group_offer}" onclick="return false"`, or use the `check_box_tag` helper

app/views/certificates/show.html.slim
```diff
@@ -34,9 +34,11 @@
...
+        = t_attr(:group_offer)
     dl
       dt= t_attr(:duration)
-      dd= "#{l(@certificate.duration_start)} #{t('to')} #{l(@certificate.duration_end) if @certificate.duration_end}"
+      dd= "#{l(@certificate.duration_start) if @certificate.duration_start} #{t('to')} #{l(@certificate.duration_end) if @certificate.duration_end}"
```
toupeira: You can also do `l ..., default: ''` to avoid the `if`

config/locales/en.yml
```diff
@@ -386,6 +386,8 @@ en:
       aprox_hours_given: 'Aprox. %{hours} Hours given'
       back_to_volunteer: Back to volunteer
     edit: &id-certificates-form
+      group_offer: Gruppenangebot
+      assignment_kinds: Assignment types
```
toupeira: Key already exists below :)

app/views/certificates/_form.html.slim
```diff
@@ -17,6 +17,7 @@
       .row
         - Assignment::KINDS.each do |kind|
           .col-xs-2= fk.input kind, as: :boolean, input_html: { checked: @certificate.assignment_kinds[kind.to_s] }
+    = f.input :group_offer, as: :boolean, checked: @certificate.group_offer.to_s
```
toupeira: Hmm where does `@certificate.group_offer` come from? If I create a certificate in the console and try to call `group_offer` on it I get a `NoMethodError`.

app/views/certificates/index.html.slim
```diff
@@ -4,8 +4,8 @@ h1= t('.volunteers_certificates', name: @volunteer.contact.full_name)
   table.table
     thead
       tr
-        th= t('created_at')
-        th= t('updated_at')
+        th= t_attr(:created_at, Certificate)
```
chrysalag: `created_at` and `updated_at` are included in the `generic_keys` so you can just have `t_attr(:created_at)` (and remove the individual translations).

config/locales/en.yml
```diff
@@ -386,6 +386,7 @@ en:
       aprox_hours_given: 'Aprox. %{hours} Hours given'
       back_to_volunteer: Back to volunteer
     edit: &id-certificates-form
+      group_offer: Gruppenangebot
```
chrysalag: english :)

app/views/certificates/_form.html.slim
nuritnt: https://github.com/panter/aoz-003/blob/121c93124308efa6086b053e7e0f46a0deff634e/app/models/certificate.rb
in `certificate model` on line 12: 
```
def build_values
......
...group_offer: volunteer.group_offers.any?)
end
hmm.. but why doesn't it throw me an error in the views when one is created?

Comment (toupeira) [reply app/views/certificates/_form.html.slim]: Yeah looks like there are a few problems with the `Certificate` model :) Can you add some unit tests for the `build_values` behavior?

```ruby
[1] pry » c = Certificate.new
Creating scope :without_assignment. Overwriting existing method Volunteer.without_assignment.
NoMethodError: undefined method `hours' for nil:NilClass
from /home/mak/src/panter/aoz-003/app/models/certificate.rb:10:in `build_values'
```

```ruby
[2] pry » c = Certificate.new volunteer: Volunteer.first
...
ActiveModel::UnknownAttributeError: unknown attribute 'group_offer' for Certificate.
from /home/mak/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/activemodel-5.1.4/lib/active_model/attribute_assignment.rb:48:in `_assign_attribute'
```

toupeira: Also, since the build passes we should probably add a feature spec to test certificate creation. I was trying to test it manually through the interface but couldn't figure out where to create them :)

app/models/certificate.rb
```diff
@@ -1,16 +1,22 @@
...
+    self.minutes ||= volunteer.hours.minutes_rest
+    self.assignment_kinds ||= volunteer.assignment_kinds
+    self.volunteer_contact ||= convert_volunteer_contact
+    self.institution ||= DEFAULT_INSTITUTION
+    self.assignments_duration ||= volunteer.assignments_duration
+    self.group_offer ||= has_group_offer
+  end
+
+  def has_group_offer
+    volunteer.group_offers = true if !nil?
```
nuritnt: @manufaktor does this implementation make sense so far? also not sure about checking the `group_offer boolean`

app/controllers/certificates_controller.rb
```diff
@@ -1,6 +1,6 @@
 class CertificatesController < ApplicationController
-  before_action :set_certificate, except: [:index, :new]
-  before_action :set_volunteer, only: [:index, :new]
+  before_action :set_certificate, except: [:show, :index, :new, :create]
```
manufaktor: `:create` seems wrong here, since `set_certificate` tries to find a Certificate which isn't created yet

app/controllers/certificates_controller.rb
```diff
@@ -20,16 +20,15 @@ def show
...
+    @certificate.build_values
     authorize @certificate
   end
 
   def edit; end
 
-  def update
-    @certificate.update(certificate_params.merge(volunteer_id: @volunteer.id,
+  def create
+    @certificate = Certificate.new(certificate_params.merge(volunteer_id: @volunteer.id,
```
manufaktor: you could probably use `volunteer: @volunteer` or directly build on the `@volunteer` association.

app/controllers/certificates_controller.rb
```diff
@@ -20,16 +20,15 @@ def show
 
   def new
     @certificate = Certificate.new(volunteer_id: @volunteer.id, user_id: current_user.id)
```
manufaktor: you could probably use volunteer: @volunteer or directly build on the @volunteer association.

app/controllers/certificates_controller.rb
```diff
@@ -38,6 +37,14 @@ def update
     end
   end
 
+  def update
+    if @certificate.update(certificate_params)
+      redirect_to volunteer_certificate_path(@volunteer, @certificate)
+    else
+      render :new
```
manufaktor: why render `new`, shoudn't this be `edit`?

app/controllers/certificates_controller.rb
```diff
@@ -47,18 +54,18 @@ def destroy
 
   def set_certificate
     @certificate = Certificate.find(params[:id])
-    @volunteer = @certificate.volunteer
```
manufaktor: this should remain, or you will not have a volunter in the show action.

app/controllers/certificates_controller.rb
```diff
@@ -47,18 +54,18 @@ def destroy
...
     authorize @certificate
   end
 
   def set_volunteer
-    @volunteer = Volunteer.find(params[:volunteer_id])
+    @volunteer = Volunteer.find(params[:volunteer_id]) if params[:volunteer_id]
```
manufaktor: shouldn't the `volunteer_id` always be present (you're in a nested route) ?

app/models/certificate.rb
```diff
@@ -1,16 +1,21 @@
 class Certificate < ApplicationRecord
-  after_initialize :build_values
-
   belongs_to :volunteer
   belongs_to :user, -> { with_deleted }
 
   def build_values
     return unless text_body.nil?
```
manufaktor: I don't think this is needed

app/models/certificate.rb
```diff
@@ -1,16 +1,21 @@
...
-      institution: DEFAULT_INSTITUTION)
-    update(volunteer.assignments_duration)
+    self.text_body ||= default_text_body
+    self.function ||= DEFAULT_FUNCTION
+    self.hours ||= volunteer.hours.total_hours
+    self.minutes ||= volunteer.hours.minutes_rest
+    self.assignment_kinds ||= volunteer.assignment_kinds
+    self.volunteer_contact ||= convert_volunteer_contact
+    self.institution ||= DEFAULT_INSTITUTION
+    self.group_offer ||= has_group_offer
```
manufaktor: this would fail if `self.group_offset` is false, and it doesn't make sense to initialize it with a `nil` value

app/controllers/certificates_controller.rb
chrysalag: Sorry, I accidentally deleted @manufaktor comment. 
The comment was:
> `:create` seems wrong here, since set_certificate tries to find a Certificate which isn't created yet

chrysalag: So this could be written `only: [:edit, :update, :destroy]`, as well

app/models/certificate.rb
```diff
@@ -1,16 +1,21 @@
...
+  end
+
+  def has_group_offer
+    true unless volunteer.group_offers.nil?
```
chrysalag: or `volunteer.group_offers.blank?`

test/system/billing_expenses_test.rb
```diff
@@ -42,4 +53,6 @@ def setup
     assert page.has_text? 'PLZ / Ort'
     assert page.has_text? 'Bank / IBAN'
   end
+
```
chrysalag: NIT: remove empty lines

test/system/billing_expenses_test.rb
```diff
@@ -4,8 +4,12 @@ class BillingExpensesTest < ApplicationSystemTestCase
   def setup
     superadmin = create :user
     @volunteer = create :volunteer
-    assignment = create :assignment, volunteer: @volunteer
-    create :hour, volunteer: @volunteer, hourable: assignment
+    @assignment = create :assignment, volunteer: @volunteer
+    @hour1 = create :hour, volunteer: @volunteer, hourable: @assignment, hours: '2', minutes: '30'
+    category = create :group_offer_category
+    @group_offer = create :group_offer, volunteers: [@volunteer], group_offer_category: category
```
chrysalag: The group offer factory calls the group offer category one, so:
`, group_offer_category: category` and the above line are not needed.

chrysalag: The group offer factory calls the group offer category one, no need to be defined.

app/views/certificates/_form.html.slim
```diff
@@ -15,7 +15,7 @@
     h4= t('.assignment_kinds')
     = f.simple_fields_for :assignment_kinds do |fk|
       .row
-        - Assignment::KINDS.each do |kind|
+        - [:group_offer, :assignment].each do |kind|
```
casaper: It just popped up here I forgot:
```slim
- @certificate.assignment_kinds.each do |kind, value|
  .col-xs-2= fk.input kind, as: :boolean, input_html: { checked: value }
```

test/models/certificate_test.rb
```diff
@@ -0,0 +1,15 @@
...
+
+class CertificateTest < ActiveSupport::TestCase
+  test 'certificate is valid' do
+    certificate = Certificate.new
+    refute certificate.valid?
+    error_msgs = { volunteer: ['must exist'], user: ['must exist'] }
+    assert_equal error_msgs, certificate.errors.messages
+    certificate.user = create :user
+    refute certificate.valid?
+    assert_equal error_msgs.except(:user), certificate.errors.messages
```
casaper: Actually line 10 and 11 can be dropped, because they're redundant

nuritnt:
@casaper 🤓 its not completely wrong, since this story is about billing expenses _and_ certificates. but luckily for the billing expenses were not a lot of changes needed.

casaper:
@nuritnt 🆗 my bad 😉

PR #263, opened 2017-10-13
Performance reports for group offer

nuritnt:
# [Story in Trello](https://trello.com/c/pnuKvcsT/17-3-kennzahlen-f%C3%BCr-gruppeneins%C3%A4tze-auch-rapportieren)

### Done for approval?

* [x] Story under test
* [ ] ~Mobile works~
* [ ] ~Seeds created~
* [ ] ~Translated~
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -25,7 +25,9 @@ class Assignment < ApplicationRecord
 
   scope :default_order, (-> { order(created_at: :desc) })
 
-  scope :created_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
+  scope :created_between, lambda { |start_date, end_date|
+    created_before(end_date).crated_after(start_date)
+  }
   scope :created_before, ->(max_time) { where('assignments.created_at < ?', max_time) }
   scope :created_after, ->(min_time) { where('assignments.created_at > ?', min_time) }
 
@@ -58,7 +60,6 @@ class Assignment < ApplicationRecord
   scope :active_between, lambda { |start_date, end_date|
     start_within(start_date..end_date)
       .or(end_within(start_date..end_date))
-      .or(no_end.start_before(start_date))
   }
 
   scope :zurich, (-> { joins(:client).merge(Client.zurich) })
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -26,7 +26,7 @@ class Assignment < ApplicationRecord
   scope :default_order, (-> { order(created_at: :desc) })
 
   scope :created_between, lambda { |start_date, end_date|
-    created_before(end_date).crated_after(start_date)
+    created_before(end_date).created_after(start_date)
   }
   scope :created_before, ->(max_time) { where('assignments.created_at < ?', max_time) }
   scope :created_after, ->(min_time) { where('assignments.created_at > ?', min_time) }
@@ -42,6 +42,7 @@ class Assignment < ApplicationRecord
   scope :end_in_future, (-> { where('period_end > ?', Time.zone.today) })
   scope :not_ended, (-> { no_end.or(end_in_future) })
 
+  scope :no_start, (-> { where(period_start: nil) })
   scope :started, (-> { where('period_start < ?', Time.zone.today) })
   scope :will_start, (-> { where('period_start > ?', Time.zone.today) })
   scope :start_before, ->(date) { where('period_start < ?', date) }
@@ -55,7 +56,7 @@ class Assignment < ApplicationRecord
   end
 
   scope :active, (-> { not_ended.started })
-  scope :inactive, (-> { ended })
+  scope :inactive, (-> { ended.or(no_start) })
 
   scope :active_between, lambda { |start_date, end_date|
     no_end.start_before(end_date)
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1,5 +1,6 @@
 class Assignment < ApplicationRecord
   include ImportRelation
+  include GroupAssignmentAndAssignmentScopes
 
   after_update :delete_reminder, if: :saved_change_to_confirmation?
 
@@ -23,45 +24,18 @@ class Assignment < ApplicationRecord
 
   validates :client_id, uniqueness: { scope: :volunteer_id, message: I18n.t('assignment_exists') }
 
-  scope :default_order, (-> { order(created_at: :desc) })
 
-  scope :created_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
   scope :created_before, ->(max_time) { where('assignments.created_at < ?', max_time) }
   scope :created_after, ->(min_time) { where('assignments.created_at > ?', min_time) }
 
-  scope :no_end, (-> { where(period_end: nil) })
-  scope :has_end, (-> { where.not(period_end: nil) })
 
   scope :ended, (-> { where('assignments.period_end < ?', Time.zone.today) })
-  scope :end_before, ->(date) { where('period_end < ?', date) }
-  scope :end_after, ->(date) { where('period_end > ?', date) }
-  scope :end_within, ->(date_range) { where(period_end: date_range) }
 
-  scope :end_in_future, (-> { where('period_end > ?', Time.zone.today) })
-  scope :not_ended, (-> { no_end.or(end_in_future) })
-
-  scope :no_start, (-> { where(period_start: nil) })
-  scope :started, (-> { where('period_start < ?', Time.zone.today) })
-  scope :will_start, (-> { where('period_start > ?', Time.zone.today) })
-  scope :start_before, ->(date) { where('period_start < ?', date) }
-  scope :start_after, ->(date) { where('period_start > ?', date) }
-  scope :start_within, ->(date_range) { where(period_start: date_range) }
-
-  scope :started_six_months_ago, (-> { where('period_start < ?', 6.months.ago.to_date) })
 
   def started_six_months_ago?
     period_start < 6.months.ago.to_date
   end
 
-  scope :active, (-> { not_ended.started })
-  scope :inactive, (-> { ended.or(no_start) })
-
-  scope :active_between, lambda { |start_date, end_date|
-    no_end.start_before(end_date)
-          .or(
-            start_before(end_date).end_after(start_date)
-          )
-  }
 
   scope :zurich, (-> { joins(:client).merge(Client.zurich) })
   scope :not_zurich, (-> { joins(:client).merge(Client.not_zurich) })
@@ -69,7 +43,6 @@ class Assignment < ApplicationRecord
   scope :internal, (-> { joins(:volunteer).merge(Volunteer.internal) })
   scope :external, (-> { joins(:volunteer).merge(Volunteer.external) })
 
-  scope :with_hours, (-> { joins(:hours) })
 
   def ended?
     period_end && period_end < Time.zone.today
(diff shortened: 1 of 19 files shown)


app/models/performance_report.rb
```diff
@@ -12,15 +12,17 @@ def generate_report
     end
     self.report_content = {
       global: global,
-      zuerich: zuerich
+      zuerich: zuerich,
+      internal_offer: internal_offer
```
casaper: the internal external thing is covered with the boolean. So either the whole report is internal or it is global

app/models/group_assignment.rb
```diff
@@ -10,6 +10,24 @@ class GroupAssignment < ApplicationRecord
...
+
+  scope :start_within, ->(date_range) { start_before(date_range.last).start_after(date_range.first) }
+  scope :end_within, ->(date_range) { where(period_end: date_range) }
+  scope :end_after, ->(date) { where('period_end > ?', date) }
+  scope :end_before, ->(date) { where('period_end < ?', date) }
+
+  scope :start_after, ->(date) { where('period_start > ?', date) }
+  scope :start_before, ->(date) { where('period_start < ?', date) }
+
+  scope :not_end_before, ->(date) { where('period_end > ?', date) }
```
chrysalag: That is the same with `:end_after` scope.

app/models/assignment.rb
```diff
@@ -53,15 +56,20 @@ def started_six_months_ago?
...
   scope :active, (-> { not_ended.started })
-  scope :inactive, (-> { ended })
+  scope :inactive, (-> { ended.or(no_start) })
 
   scope :active_between, lambda { |start_date, end_date|
-    start_within(start_date..end_date)
-      .or(end_within(start_date..end_date))
-      .or(no_end.start_before(start_date))
+    no_end.start_before(end_date)
+          .or(
```
chrysalag: Could it also need `no_start.end_after(start_date)`?

app/models/group_assignment.rb
```diff
@@ -10,6 +10,24 @@ class GroupAssignment < ApplicationRecord
...
+
+  scope :active_between, lambda { |start_date, end_date|
+    no_end.start_before(end_date)
+          .or(
```
chrysalag: Could it also need `no_start.end_after(start_date)`? If yes, `:no_start` scope is not defined yet.

app/models/group_offer.rb
```diff
@@ -25,6 +25,35 @@ class GroupOffer < ApplicationRecord
...
+
+  def has_active_assignments_between?(start_date, end_date)
+    group_assignments.active_between(start_date, end_date).size.positive?
+  end
+
+  scope :created_before, ->(date) { where('created_at < ?', date) }
+
+  def all_group_assignments_ended_within?(date_range)
+    # TODO
+    # are group assignments put in group_assignments_log on group_offer.destroy ?
```
chrysalag: Yes, that was fixed at #283.

app/models/group_offer.rb
```diff
@@ -25,6 +25,35 @@ class GroupOffer < ApplicationRecord
...
+    not_end_before += group_assignments.no_end.ids if date_range.last >= Time.zone.today
+    ended_within.size.positive? && not_end_before.blank?
+  end
+
+  def all_group_assignments_started_within?(date_range)
+    # TODO
+    # are group assignments put in group_assignments_log on group_offer.destroy ?
+    started_within = group_assignments.start_within(date_range)
+    started_before = group_assignments.start_before(date_range.first)
+    return true if started_within.size == group_assignments.size
```
chrysalag: Or maybe `started_within.size == group_assignments.size || started_before.size.zero?`?

app/models/group_offer.rb
```diff
@@ -25,6 +25,35 @@ class GroupOffer < ApplicationRecord
 
   scope :in_department, (-> { where.not(department_id: nil) })
 
+  scope :active_group_assignments_between, lambda { |start_date, end_date|
+    joins(:group_assignments).merge(GroupAssignment.active_between(start_date, end_date))
+  }
+
+  def has_active_assignments_between?(start_date, end_date)
```
chrysalag: Maybe rename that to `has_active_group_assignments_between?`

app/models/volunteer.rb
```diff
@@ -70,31 +70,43 @@ class Volunteer < ApplicationRecord
...
-  scope :with_active_assignments, (-> { joins(:assignments).merge(Assignment.active) })
   scope :with_active_assignments_between, lambda { |start_date, end_date|
     joins(:assignments).merge(Assignment.active_between(start_date, end_date))
   }
+  scope :with_active_group_assignments_between, lambda { |start_date, end_date|
+    joins(:group_assignments).merge(GroupAssignment.active_between(start_date, end_date))
+  }
+
+  scope :with_active_assignments, (-> { joins(:assignments).merge(Assignment.active) })
+  scope :with_active_group_offers, (-> { joins(:group_offers).merge(GroupOffer.active) })
```
chrysalag: It is possible though that a volunteer is removed from any group offer they were in and the group offer(s) remain(s) active.

test/models/group_assignment_scopes_test.rb
```diff
@@ -0,0 +1,139 @@
...
+  test 'start_within' do
+    started_before = create_group_assignments 100.days.ago, nil, create(:volunteer)
+    started_before_ended_within = create_group_assignments 100.days.ago, 40.days.ago, create(:volunteer)
+    started_after = create_group_assignments 10.days.ago, nil, create(:volunteer)
+    started_after_ended = create_group_assignments 20.days.ago, 10.days.ago, create(:volunteer)
+    started_within_no_end = create_group_assignments 40.days.ago, nil, create(:volunteer)
+    started_within_ended_within = create_group_assignments 40.days.ago, 35.days.ago, create(:volunteer)
+    started_within_ended_after = create_group_assignments 40.days.ago, 20.days.ago, create(:volunteer)
+    query = GroupAssignment.start_within(45.days.ago..30.days.ago)
+    assert query.include? started_within_no_end.first
```
chrysalag: Why `.first`?

test/models/group_offer_scopes_test.rb
```diff
@@ -0,0 +1,101 @@
+require 'test_helper'
+
+class GroupOfferScopesTest < ActiveSupport::TestCase
+  test 'active' do
+    group_offer_active = create :group_offer, active: true
```
chrysalag: It is true by default

app/models/assignment.rb
casaper: No, because a assignment with no start is not, and was never active

app/models/group_assignment.rb
casaper: Same as above. No start means never active

app/models/group_offer.rb
casaper: For the sake of explicity yes, but don't blame me for rubocop hassling for too long lines, if only the method name without parameters fills 50% of the allowed line lengths 😉

test/models/group_assignment_scopes_test.rb
casaper: it returns an array. thats why. needs to be like that

test/models/group_offer_scopes_test.rb
casaper: who cares about defaults. Where testing here, so what if the default is ever changed?

app/models/assignment.rb
```diff
@@ -25,7 +25,9 @@ class Assignment < ApplicationRecord
 
   scope :default_order, (-> { order(created_at: :desc) })
 
-  scope :created_between, ->(start_date, end_date) { where(created_at: start_date..end_date) }
+  scope :created_between, lambda { |start_date, end_date|
```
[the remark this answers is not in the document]
toupeira: Not sure, looks like the previous version used `BETWEEN` in the SQL (same as `<=` / `>=`) while the new one uses `<` / `>`, so the behaviour is slightly different.

app/models/group_assignment.rb
```diff
@@ -10,6 +10,22 @@ class GroupAssignment < ApplicationRecord
...
+
+  scope :start_after, ->(date) { where('period_start > ?', date) }
+  scope :start_before, ->(date) { where('period_start < ?', date) }
+
+  scope :active_between, lambda { |start_date, end_date|
+    no_end.start_before(end_date)
+          .or(
+            start_before(end_date).end_after(start_date)
+          )
+  }
```
toupeira: Maybe move these scopes into a new concern which could also be used in `Assignment`?

app/models/group_offer.rb
```diff
@@ -25,6 +25,31 @@ class GroupOffer < ApplicationRecord
 
   scope :in_department, (-> { where.not(department_id: nil) })
 
+  scope :active_group_assignments_between, lambda { |start_date, end_date|
+    joins(:group_assignments).merge(GroupAssignment.active_between(start_date, end_date))
+  }
+
+  def has_active_group_assignments_between?(start_date, end_date)
+    group_assignments.active_between(start_date, end_date).size.positive?
```
toupeira: Use `.exists?` instead of `.size.positive?`

app/models/group_offer.rb
```diff
@@ -25,6 +25,31 @@ class GroupOffer < ApplicationRecord
...
+  end
+
+  scope :created_before, ->(date) { where('created_at < ?', date) }
+
+  def all_group_assignments_ended_within?(date_range)
+    ended_within = group_assignments.end_within(date_range).ids
+    not_end_before = group_assignments.end_after(date_range.last).ids
+    not_end_before += group_assignments.no_end.ids if date_range.last >= Time.zone.today
+    ended_within.size.positive? && not_end_before.blank?
```
toupeira: Use `.exists?` instead of `.size.positive?`

app/models/group_offer.rb
```diff
@@ -25,6 +25,31 @@ class GroupOffer < ApplicationRecord
...
+  end
+
+  def all_group_assignments_started_within?(date_range)
+    started_within = group_assignments.start_within(date_range)
+    started_before = group_assignments.start_before(date_range.first)
+    return true if started_within.size == group_assignments.size
+    return true if started_before.size.zero?
```
toupeira: Use `!started_before.exists?` instead of `.size.zero?`

app/models/performance_report.rb
```diff
@@ -55,12 +58,32 @@ def clients(zurich = false)
...
+      group_offer.has_active_group_assignments_between?(period_start, period_end)
+    end
+    ended = group_offers.map do |group_offer|
+      group_offer.all_group_assignments_ended_within?(period_start..period_end)
+    end
+    new_group_offers = group_offers.map do |group_offer|
+      group_offer.all_group_assignments_started_within?(period_start..period_end)
+    end
+    {
+      active: active.reject(&:blank?).size,
```
toupeira: Try `.compact` instead of `.reject(&:blank)`, should work the same if the blank values are `nil`

app/models/group_offer.rb
casaper: Your right, but how about `any?`

app/models/performance_report.rb
casaper: but they are not nil, they're false

casaper: nope, I just checked in pry:
```ruby
[false, false, true].compact # => [ [0] false, [1] false, [2] true ]
```

casaper: but as I just discovered, `.reject(&:blank?)` only works in rails, not in plain ruby.

casaper: but maybe 
```ruby
active.grep(true)
```
?

PR #268, opened 2017-10-18
Features/group offers index print

chrysalag:
# [Story in Trello](https://trello.com/c/d3UUP5JT/40-ausdruck-%C3%BCbersicht-gruppenangebote)

### Done for approval?

* [x] Story under test
* ~[ ] Mobile works~
* ~[ ] Seeds created~
* [x] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/views/group_offers/index.xlsx.axlsx
+++ b/app/views/group_offers/index.xlsx.axlsx
@@ -1,8 +1,10 @@
 wb = xlsx_package.workbook
 
-col_header = wb.styles.add_style(:bg_color => "FFDFDEDF", b: true , alignment: { horizontal: :center, vertical: :center }, border: {
-  color: '00', edges: [:bottom], style: :thin
-})
+col_header = wb.styles.add_style(
+  bg_color: 'FFDFDEDF',
+  b: true,
+  alignment: { horizontal: :center, vertical: :center },
+  border: { color: '00', edges: [:bottom], style: :thin})
 
 wb.add_worksheet(name: t('group_offers', count: 2)) do |sheet|
   header_row = [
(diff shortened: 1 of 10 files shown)


app/views/group_offers/index.xlsx.axlsx
```diff
@@ -0,0 +1,72 @@
...
+    end
+    sheet.add_row [
+      offer.title,
+      location,
+      availabilities,
+      targets,
+      durations,
+      state,
+      volunteers,
+      offer.group_offer_category
```
nuritnt: screen shot 2017-10-18 at 12 58 54

`duration`cell is missing, thats why the content of the cells are shifted:  `volunteers` are in the `category cell` displayed

app/views/group_offers/index.xlsx.axlsx
```diff
@@ -0,0 +1,72 @@
...
+          targets << t_attr(target)
+          targets << ", "
+        end
+      end
+    end
+    durations = ""
+    GroupOffer::DURATION.each do |duration|
+      if offer[duration]
+        durations << t_attr(duration)
+        durations << " "
```
nuritnt: screen shot 2017-10-18 at 13 05 17

screen shot 2017-10-18 at 12 58 54


`duration`cell is not shown eventough in one group offer its existing, thats why the content of the cells are shifted:  `volunteers` are in the `category cell` displayed

test/integration/group_offers_export_test.rb
```diff
@@ -0,0 +1,15 @@
...
+  def setup
+    @superadmin = create :user
+    10.times { create :group_offer }
+  end
+
+  test 'xlsx file is downloadable' do
+    login_as @superadmin
+    get group_offers_url(format: :xlsx)
+    assert_equal 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+      response.content_type
```
nuritnt: how easy is it to include in the test if the correct fields are shown in the excel file?

app/views/group_offers/index.xlsx.axlsx
chrysalag: True, I forgot to add it as a column.

chrysalag: Added

app/views/group_offers/index.xlsx.axlsx
```diff
@@ -0,0 +1,73 @@
+wb = xlsx_package.workbook
+
+col_header = wb.styles.add_style(:bg_color => "FFDFDEDF", b: true , alignment: { horizontal: :center, vertical: :center }, border: {
```
toupeira: Use newer hash syntax `bg_color:`

Also IMHO it would look nicer to split it into more lines, i.e.

```ruby
col_header = wb.styles.add_style(
  bg_color: ...
  ...
)
```

app/views/group_offers/index.xlsx.axlsx
```diff
@@ -0,0 +1,73 @@
...
+    t_attr(:group_offer_category)
+  ]
+
+  sheet.add_row header_row, style: col_header
+
+  cell_types = Array.new(header_row.size).map { :string }
+  cell_styles = Array.new(header_row.size).map { nil }
+  @group_offers.each do |offer|
+    location = "#{offer.organization} #{offer.location}" if offer.external?
+    location = "#{offer.department.contact.last_name}" if !offer.external? && !offer.department.nil?
```
toupeira: I would combine the conditions:

```ruby
location =
  if offer.external?
    "#{offer.organization} #{offer.location}"
  elsif offer.department
    "#{offer.department.contact.last_name}"
  else
    # what should happen in this case?
  end
```

Maybe it would also make sense to move this into a method on `GroupOffer`

app/views/group_offers/index.xlsx.axlsx
```diff
@@ -0,0 +1,73 @@
...
+    availabilities = ""
+    availability_collection.each do |availability|
+      if offer[availability]
+        availabilities << t("availability.#{availability}")
+        availabilities << ", "
+      end
+    end
```
toupeira: Not sure if the trailing comma is desired, you could also do something like:

```ruby
availabilities = availability_collection.map do |availability|
  t("availability.#{availability}") if offer[availability]
end.compact.join(', ')
```

toupeira: And similarly for the other loops below

test/integration/group_offers_export_test.rb
chrysalag: Not so sure, I haven't managed it yet. Also I am not sure if it is needed, for example we don't check for the index columns either and we didn't think of testing the xlsx fields in the client and volunteers.

toupeira: Looks like the `axlsx` gem doesn't support reading Excel files, but there are [a few other gems](https://spin.atomicobject.com/2017/03/22/parsing-excel-files-ruby/).

You can get the file contents from `response.body`, if the gem only supports loading files you can use `Tempfile` from Ruby's standard library:

```ruby
excel_file = Tempfile.new
excel_file.write(response.body)
excel_file.close
Roo::Spreadsheet.open(excel_file.path)
```

The nice thing about `Tempfile` is that the file will be automatically deleted once the variable goes out of scope (i.e. after the test finishes)

toupeira: > Also I am not sure if it is needed

IMHO it's definitely low-priority :)

nuritnt: yup agree - its not really priority, just came in my mind, when I noticed a cell was missing in the excel

PR #270, opened 2017-10-18
fix: assignment_hours

nuritnt:
hours were only available to work if there was an assignment, but not working only if group offer is existing. 
now it is also checking for entering the hours, if the assignments even started.

* [ ] Story under test
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/volunteer.rb
+++ b/app/models/volunteer.rb
@@ -157,6 +157,14 @@ class Volunteer < ApplicationRecord
     assignments.size.positive?
   end
 
+  def assignment_started?
+    assignments.started.size.positive?
+  end
+
+  def group_assignment_stared?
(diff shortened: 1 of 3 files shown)


app/models/volunteer.rb
```diff
@@ -165,8 +173,8 @@ def internal?
     !external
   end
 
-  def internal_and_assignments?
-    internal? && assignments?
+  def internal_and_started_assignments?
+    internal? && assignment_started? || group_assignment_stared?
```
chrysalag: `internal? && (assignment_started? || group_assignment_stared?)`
`&&` [higher priority](has http://www.techotopia.com/index.php/Ruby_Operator_Precedence) than `||` so it executes first.

PR #282, opened 2017-10-19
WIP: Fix seeking clients page and scopes

chrysalag:
# [Story in Trello](https://trello.com/c/i0kLXa0C/63-bug-https-stagingaoz-freiwilligech-volunteers-seekingclients-doesnt-work)

### Done for approval?

* [ ] Story under test
* ~[ ] Mobile works~
* ~[ ] Seeds created~
* ~[ ] Translated~
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/volunteer.rb
+++ b/app/models/volunteer.rb
@@ -83,7 +83,7 @@ class Volunteer < ApplicationRecord
   scope :without_assignment, (-> { left_outer_joins(:assignments).where(assignments: { id: nil }) })
   scope :without_active_assignment, (-> { joins(:assignments).merge(Assignment.ended) })
   scope :not_in_any_group_offer, lambda {
-    left_joins(:group_offers).where(group_assignments: { volunteer_id: nil })
+    left_outer_joins(:group_assignments).where(group_assignments: { id: nil })
   }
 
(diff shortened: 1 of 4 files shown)


app/models/volunteer.rb
```diff
@@ -114,17 +114,29 @@ class Volunteer < ApplicationRecord
       .merge(Assignment.inactive)
       .where.not(assignments: { volunteer_id: with_active_assignments.ids })
       .or(loj_without_assignments)
+      .merge(
+        left_outer_joins(:group_assignments)
+          .where('group_assignments.period_end IS NULL
+            OR group_assignments.period_end < ?', Time.zone.today)
```
nuritnt: can't you use here the scopes where `started`, `ongoing`, `no_end` are already defined in the `group assignment` [model](https://github.com/panter/aoz-003/blob/develop/app/models/group_assignment.rb)? or was this the problem we mentioned in todays standup?

casaper: @nuritnt no, there not compatible, because they use different SQL joins `LEFT OUTER JOIN`, `LEFT JOIN`, etc.. aren't compatible

PR #308, opened 2017-10-27
Bugfix/group offer volunteers edit form as department manager not working

casaper:
# [Story in Trello](https://trello.com/c/QEk7bZuv/21-bug-group-offer-edit-existing-nested-attribute-volunteers-fields-are-displayed-but-not-filled-with-data-so-on-submit-the-form-is)

### Done for approval?

* [x] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [ ] Translated
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/group_assignment.rb
+++ b/app/models/group_assignment.rb
@@ -2,7 +2,6 @@ class GroupAssignment < ApplicationRecord
   include GroupAssignmentAndAssignmentScopes
 
   belongs_to :group_offer
-  validates_presence_of :group_offer
 
   belongs_to :volunteer
   has_many :group_assignment_logs
--- a/app/models/group_assignment.rb
+++ b/app/models/group_assignment.rb
@@ -2,11 +2,15 @@ class GroupAssignment < ApplicationRecord
   include GroupAssignmentAndAssignmentScopes
 
   belongs_to :group_offer
+  validates_presence_of :group_offer
+
   belongs_to :volunteer
   has_many :group_assignment_logs
 
   after_update :save_group_assignment_logs, if: :dates_updated?
   before_destroy :save_group_assignment_logs
+
+  validates_uniqueness_of :volunteer, scope: :group_offer
 
   scope :ongoing, (-> { where('group_assignments.period_end > ?', Time.zone.today) })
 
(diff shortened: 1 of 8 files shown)


db/migrate/20171026145935_group_assignment_uniqueness_group_offer_combined_with_volunteer.rb
```diff
@@ -0,0 +1,7 @@
+class GroupAssignmentUniquenessGroupOfferCombinedWithVolunteer < ActiveRecord::Migration[5.1]
+  def change
+    add_column :group_assignments, :active, :boolean, default: true
```
nuritnt: for what is this needed in the group_assignments? isn't it already handled in the group offer itself?

casaper: That is the recommended way, to do it as well in the db.

app/models/group_assignment.rb
```diff
@@ -2,12 +2,16 @@ class GroupAssignment < ApplicationRecord
   include GroupAssignmentAndAssignmentScopes
 
   belongs_to :group_offer
+  validates_presence_of :group_offer
```
chrysalag: Why is this [needed](https://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html)?

db/migrate/20171026145935_group_assignment_uniqueness_group_offer_combined_with_volunteer.rb
chrysalag: What is meant by active?

casaper:
@chrysalag 
> What is meant by active?

This is working with paranoia gem, same as with the Users table. This way only not softdeleted records are checked for uniqueness. deleted records are not throwing an error if they'd duplicate with a not soft deleted one

db/migrate/20171026145935_group_assignment_uniqueness_group_offer_combined_with_volunteer.rb
casaper: This is working with paranoia gem, same as with the Users table. This way only not softdeleted records are checked for uniqueness. deleted records are not throwing an error if they'd duplicate with a not soft deleted one

casaper: the field name is given by paranoiagem

app/models/group_assignment.rb
casaper: its not, its rubbel

PR #312, opened 2017-10-27
Features/modifications to client index

casaper:
# [Story in Trello](https://trello.com/c/xeDf9HZE/31-klienten-index-anpassungen)

### Done for approval?

* [x] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [ ] Translated
* [x] Code is reviewed

### Done after the merge?


app/views/application/_index_actions.html.slim
```diff
@@ -1,10 +1,5 @@
...
-  td= link_to t_model(Journal), { controller: 'journals', action: 'index', "#{target_controller.singularize}_id": subject.id }
-- if policy(subject.class).index?
-  td= link_to t_action(:edit), { controller: target_controller, action: :edit, id: subject.id }
-- if policy(subject.class).destroy?
-  td= link_to t_action(:destroy), subject, confirm_deleting(subject)
+td.index-action-cell
+  span= link_to t_action(:show), subject
+  span= link_to t_action(:edit), edit_polymorphic_path(subject)
+  - if policy(subject.class).destroy?
+    span= link_to t_action(:destroy), subject, confirm_deleting(subject)
```
toupeira: The `target_controller ||= controller_name` was already removed once and then added back to fix a bug, are you sure you're not introducing that bug again? It's still referenced in `app/views/assignments/_assignment.html.slim`:

```slim
= render 'index_actions', subject: assignment, target_controller: 'assignments'
```

app/views/assignments/find_client.html.slim
```diff
@@ -22,7 +20,57 @@ h3= t_attr(:single_accompaniment, Volunteer)
...
+h1= t('clients_need_accompanying')
+
+nav.navbar.section-navigation
+  ul.list-inline
+    li= render 'gender_request_select'
+    li= render 'age_request_select'
+    li= render 'language_skills_language_select'
+
+.text-center
+  = will_paginate @need_accompanying, renderer: BootstrapPagination::Rails
```
toupeira: We could add a helper (`paginate` or maybe `bootstrap_paginate`) to avoid having to specify the `renderer:` every time.

app/views/assignments/find_client.html.slim
```diff
@@ -22,7 +20,57 @@ h3= t_attr(:single_accompaniment, Volunteer)
...
+        th= sort_link @q, :gender_request, t_attr(:gender_request, Client)
+        th= sort_link @q, :age_request, t_attr(:age_request, Client)
+        th colspan='2'
+    tbody
+      - @need_accompanying.each do |client|
+        tr
+          td = t("salutation.#{client.salutation}") if client.salutation?
+          td = client.contact.full_name
+          td = client.contact.full_address
+          td = client.birth_year.year if client.birth_year
```
toupeira: `client.birth_year&.year`

app/views/assignments/find_client.html.slim
```diff
@@ -22,7 +20,57 @@ h3= t_attr(:single_accompaniment, Volunteer)
...
+          td
+            - client.relatives.each do |relative|
+              = relative.to_s
```
toupeira: I think `to_s` is called by default already

app/views/clients/_client.html.slim
```diff
@@ -1,30 +1,25 @@
 tr
-  td = t("salutation.#{client.salutation}") if client.salutation?
-  td = client.contact.full_name
-  td = client.contact.full_address
-  td = client.birth_year.year if client.birth_year
+  td= t("salutation.#{client.salutation}") if client.salutation?
+  td= client.contact.full_name
+  td= client.contact.city
+  td= client.contact.postal_code
+  td= client.birth_year.year if client.birth_year
```
toupeira: `client.birth_year&.year`

app/views/clients/index.html.slim
```diff
@@ -2,19 +2,17 @@ h1= t('clients')
 nav.navbar.section-navigation
   hr
   ul.list-inline
-    li= button_link t_title(:new), new_client_path, 'success', dimension: 'sm'
-    - if policy(Client).destroy?
-      li= button_link t('need_accompanying'), need_accompanying_clients_url, dimension: 'sm'
+    li= button_link 'Klient erfassen', new_client_path, dimension: 'sm'
```
toupeira: Use translation for `'Klient erfassen'`?

app/views/clients/index.html.slim
```diff
@@ -2,19 +2,17 @@ h1= t('clients')
...
+    = list_filter_dropdown(:salutation, Client.salutation_collection)
+    = list_filter_dropdown(:state, Client::STATES)
+    li= button_link t('clear_filters'), clients_path, dimension: 'sm'
+    li= button_link t('export_xlsx'), url_for(format: :xlsx, q: search_parameters), dimension: 'sm'
 
+    - if policy(Client).need_accompanying?
+      li= button_link 'Klienten ohne Begleitung', need_accompanying_clients_url, dimension: 'sm'
```
toupeira: Same here, or is this whole page only available in German?

test/system/journals_test.rb
```diff
@@ -28,22 +28,20 @@ def setup
   test 'can create journal entry by link button in show and then delete it' do
     login_as @user
     visit client_path(@client)
-    click_button 'Journal'
-    click_link 'Write a journal entry'
+    first('a', text: 'Journal').click
```
toupeira: Does `click_link 'Journal'` not work here? (also two times further down)

test/system/socialworker_test.rb
```diff
@@ -31,8 +31,7 @@ def setup
     @socialworker.clients.each do |client|
       assert page.has_text? client.contact.first_name
       assert page.has_text? client.contact.last_name
-      assert page.has_link? href: client_path(client.id)
-      assert page.has_link? href: edit_client_path(client.id)
+      assert find("a[href='#{client_path(client.id)}']")
```
toupeira: Does `page.has_link? href: client_path(client.id)` not work here?

app/views/application/_index_actions.html.slim
toupeira: See bugfix from @chrysalag in 9f01cf9d

casaper: but I think using polymorphic path would make use of that controller_name or similar obsolete ... or wouldn't it?

app/views/assignments/find_client.html.slim
casaper: Would actually be out of scope, but as it is too good to be not done, I did it.

app/views/clients/index.html.slim
casaper: It was a team decision to drop translating the application from now on, as it is an unneccesary overhead. (the application is way to specific to AOZ, so it is quite unlikeley that ever anyone would use this in english, not doing a complete refactor/self customization of the application)

app/views/clients/index.html.slim
casaper: Yes

test/system/journals_test.rb
casaper: It doesn't because there is two of these links now. (Menu on top and bottom of page)

test/system/socialworker_test.rb
casaper: That I will try. Thx

app/views/application/_index_actions.html.slim
casaper: The thing is, that problem was mainly because the actions `assignments#find_volunteer` and  `assignments#find_client` where woven tightly into the client and volunteers views. I just untangled this with clients for now, and Intend to to the same with Volunteers views ASAP, because those views are unreadable and generally just error prone.

toupeira: Hmm not sure, the test from the bugfix still seems to pass so I guess it's fine? :)

app/views/clients/index.html.slim
toupeira: Ok noted!

test/system/journals_test.rb
toupeira: Ok makes sense

PR #313, opened 2017-10-27
Changes: Journals

chrysalag:
# [Story in Trello](https://trello.com/c/VgNoTESl/38-journal)

### Done for approval?

* [x] Story under test
* ~[ ] Mobile works~
* ~[ ] Seeds created~
* ~[ ] Translated~
* [ ] Code is reviewed

### Done after the merge?


app/controllers/journals_controller.rb
```diff
@@ -51,8 +48,12 @@ def journal_relations
   end
 
   def set_journal
-    @journal = Journal.find(params[:id])
-    authorize @journal
+    if Journal.exists?(params[:id])
```
casaper: I Think you can use find_by_id, because that won't throw an exception if id is not found

```ruby
@journal = Journal.find_by_id(params[:id])
if @journal
  authorize @journal
else
  redirect_to ...
end
```

lib/access_import/journal_transform.rb
```diff
@@ -19,6 +19,6 @@ def prepare_attributes(journal, person, assignment, user)
     2 => :conversation,
     3 => :email,
     4 => :feedback,
-    5 => :file
+    5 => :feedback
```
casaper: Is it intended that way?
```ruby
4 => :feedback,
5 => :feedback
```

chrysalag: Yes. They asked this category to be removed. So during the import the entries of this category should be mapped to another one.

app/controllers/journals_controller.rb
```diff
@@ -51,8 +48,12 @@ def journal_relations
   end
 
   def set_journal
-    @journal = Journal.find(params[:id])
-    authorize @journal
+    @journal = Journal.find_by(id: params[:id])
+    if @journal
+      authorize @journal
+    else
+      redirect_to set_journaled, notice: t('crud.c_action.destroy', model: t_model)
```
toupeira: Looks like `set_journaled` is called before each action, so you can just do `redirect_to @journaled, ...`

chrysalag: For some reason, it is not called. I get `Cannot redirect to nil!` error. Maybe because it is a non route action?

toupeira: Ah sorry didn't realize this was inside `set_journal` :) It might work if you move `before_action :set_journaled` up so it's before `:set_journal`, they should be called in the order they're defined.

app/views/journals/index.html.slim
```diff
@@ -6,24 +6,18 @@ h1= t_title(:index, Journal)
...
-              td= record.body&.truncate(40)
+              td= word_wrap(record.body)
               td= link_to record.user.full_name, record.user.profile
-              td= link_to t_action(:show), url_for([@journaled, record])
               td= link_to t_action(:edit), url_for([@journaled, record]) + '/edit'
-              td= link_to t_action(:destroy), url_for([@journaled, record]), confirm_deleting(record)
 .row
-  .col-xs-12
-= form_navigation_btn :new
+  .col-xs-12= form_navigation_btn :new
```
casaper: `form_navigation_btn` allready creates the `div.row > div.col-xs-12`, so you should use `= form_navigation_btn :new` without them.

https://github.com/panter/aoz-003/blob/develop/app/helpers/application_helper.rb#L25

lib/access_import/journal_transform.rb
casaper: ah, I see. 👍

db/migrate/20171101081636_update_journal_category.rb
```diff
@@ -0,0 +1,7 @@
+class UpdateJournalCategory < ActiveRecord::Migration[5.1]
+  def change
+    Journal.where(category: 'file').each do |journal|
+      journal.update(category: 'feedback')
+    end
```
toupeira: You can also do `Journal.where(category: 'file').update_all(category: 'feedback')` to run everything in one query, but this will skip validations etc. so just FYI :)

PR #314, opened 2017-10-31
Department manager can see own Group offers AKA validations for departmentmanager -> GroupOffer

casaper:
…rtment manager

# [Story in Trello](https://trello.com/c/U2RbOCGv/12-bug-departmentmanager-kann-groupoffer-den-er-selbst-erstellt-hat-nicht-in-index-sehen)

### Done for approval?

* [x] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [ ] Translated
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/group_offers_controller.rb
+++ b/app/controllers/group_offers_controller.rb
@@ -25,7 +25,7 @@ class GroupOffersController < ApplicationController
   end
 
   def new
-    @group_offer = GroupOffer.new
+    @group_offer = GroupOffer.new(user: current_user)
     authorize @group_offer
   end
 
@@ -33,7 +33,7 @@ class GroupOffersController < ApplicationController
 
   def create
     @group_offer = GroupOffer.new(group_offer_params)
-    @group_offer.department = current_user.department.first if current_user.manages_department?
+    @group_offer.department ||= current_user.department.first if current_user.department_manager?
     authorize @group_offer
     if @group_offer.save
       redirect_to @group_offer, make_notice
@@ -97,6 +97,6 @@ class GroupOffersController < ApplicationController
       :necessary_volunteers, :description, :women, :men, :children, :teenagers, :unaccompanied,
       :all, :long_term, :regular, :short_term, :workday, :weekend, :morning, :afternoon, :evening,
       :flexible, :schedule_details, :department_id, :organization, :location,
-      :group_offer_category_id, group_assignments_attributes)
+      :group_offer_category_id, :user_id, group_assignments_attributes)
   end
 end
--- a/app/controllers/group_offers_controller.rb
+++ b/app/controllers/group_offers_controller.rb
@@ -25,7 +25,8 @@ class GroupOffersController < ApplicationController
   end
 
   def new
-    @group_offer = GroupOffer.new(user: current_user)
+    @group_offer = GroupOffer.new(creator: current_user)
+    @group_offer.department = current_user.department.first if current_user.department_manager?
     authorize @group_offer
   end
 
@@ -33,6 +34,7 @@ class GroupOffersController < ApplicationController
 
   def create
     @group_offer = GroupOffer.new(group_offer_params)
+    @group_offer.creator = current_user
     @group_offer.department ||= current_user.department.first if current_user.department_manager?
     authorize @group_offer
     if @group_offer.save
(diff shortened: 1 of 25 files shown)


app/models/group_offer.rb
```diff
@@ -8,6 +8,7 @@ class GroupOffer < ApplicationRecord
 
   belongs_to :department, optional: true
   belongs_to :group_offer_category
+  belongs_to :user # creator
```
toupeira: Should we call the association `creator` like in `Assignment`?

app/models/group_offer.rb
```diff
@@ -83,4 +88,15 @@ def volunteers_with_roles
...
   end
+
+  private
+
+  def department_manager_has_department?
+    if user.department.blank?
+      errors.add(:user_no_department, "#{I18n.t('role.department_manager')} müssen einem Standort zugeteilt sein, "\
+        "bevor sie #{I18n.t('group_offers', count: 2)} erfassen können.")
+    elsif !user.department.include?(department)
+      errors.add(:user_wrong_department, 'Nicht der richtige Standort.')
```
toupeira: Do we need translations for the errors? You can use [string interpolation](http://www.rubydoc.info/gems/i18n/I18n/Base#translate-instance_method) for the dynamic parts.

toupeira: Also you should probably do `errors.add(:department, ...`, doesn't make much of a difference but the first argument is supposed to be the attribute with the error.

app/views/group_offers/_form.html.slim
```diff
@@ -1,6 +1,8 @@
 = simple_form_for(@group_offer) do |f|
   = simple_error_notice f
-
+  = f.hidden_field :user_id
```
toupeira: It would be better to assign the user in `#create` and not rely on (possibly unsafe) values sent by the client.

config/locales/de.yml
```diff
@@ -470,7 +470,10 @@ de:
             avatar:
               spoofed_media_type: Die Inhalte entsprechen nicht ihrer Deklaration
             avatar_content_type:
-              invalid: ist ungültig
+              invalid: ist ungültigs
```
toupeira: Remove trailing `s`

test/system/group_offers_test.rb
```diff
@@ -185,4 +185,14 @@ def setup
     assert select_values.include? volunteer_one.id
     assert select_values.include? volunteer_two.id
   end
+
+  test 'department_manager cannot access group offer pages unless there is a department assigned' do
+    department_manager = create :user, role: 'department_manager'
```
toupeira: `create :department_manager`

app/models/group_offer.rb
```diff
@@ -17,8 +18,12 @@ class GroupOffer < ApplicationRecord
 
   has_many :feedbacks, as: :feedbackable, dependent: :destroy
 
+  delegate :department_manager?, to: :user
+
   validates :title, presence: true
   validates :necessary_volunteers, numericality: { greater_than: 0 }, allow_nil: true
+  validate :department_manager_has_department?, if: :department_manager?
+  validates :department, presence: true, if: :department_manager?
```
toupeira: Since the department is required we should add this to the `:department_manager` factory, and also remove the `:with_departments` trait if other users can't have departments (not sure)

app/controllers/group_offers_controller.rb
```diff
@@ -25,15 +25,15 @@ def show
...
+    @group_offer = GroupOffer.new(user: current_user)
     authorize @group_offer
   end
 
   def edit; end
 
   def create
     @group_offer = GroupOffer.new(group_offer_params)
-    @group_offer.department = current_user.department.first if current_user.manages_department?
+    @group_offer.department ||= current_user.department.first if current_user.department_manager?
```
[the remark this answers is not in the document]
casaper: from GroupOfferParams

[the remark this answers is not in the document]
casaper: from `group_offer_params`

test/system/group_offers_test.rb
casaper: not here, because `create :department_manager` creates a user that has allready a department associated

app/models/group_offer.rb
casaper: has been discussed

db/seeds.rb
```diff
@@ -80,9 +80,9 @@ def make_lang_skills
 def create_two_group_offers(group_offer_category)
   [
     FactoryGirl.create(:group_offer, necessary_volunteers: 2,
-      group_offer_category: group_offer_category,
+      group_offer_category: group_offer_category, creator: User.find_by(role: 'department_manager'),
```
chrysalag: Why do group offer categories need creator? And how users are associated with categories? (Sorry if it is obvious)

casaper: Is not a `group_offer_category` that is created here, it's `group_offer` and group_offer_category is passed to it

chrysalag: True! Sorry!

PR #315, opened 2017-10-31
Language Skills special for Client

casaper:
# [Story in Trello](https://trello.com/c/1IRY131r/39-language-skills)

### Done for approval?

* [x] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [ ] Translated
* [ ] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -74,6 +74,6 @@ class Client < ApplicationRecord
   end
 
   def german_missing?
-    language_skills.blank? || language_skills.german.blank?
+    language_skills.german.blank?
   end
 end
--- a/app/models/client.rb
+++ b/app/models/client.rb
@@ -18,7 +18,7 @@ class Client < ApplicationRecord
   has_one :assignment
   has_one :volunteer, through: :assignments
 
-  has_one :contact, as: :contactable
+  has_one :contact, as: :contactable, dependent: :destroy
   accepts_nested_attributes_for :contact
 
   has_many :relatives, as: :relativeable, dependent: :destroy
@@ -71,5 +71,9 @@ class Client < ApplicationRecord
 
   def self.state_collection
     STATES.map(&:to_sym)
+  end
+
+  def german_missing?
+    language_skills.blank? || language_skills.german.blank?
   end
 end
--- a/app/policies/client_policy.rb
+++ b/app/policies/client_policy.rb
@@ -20,8 +20,9 @@ class ClientPolicy < ApplicationPolicy
   alias_method :need_accompanying?, :superadmin_or_department_manager?
 
   # suplementary policies
-  alias_method :supervisor?, :superadmin?
-  alias_method :journals?,   :superadmin?
-  alias_method :comments?,   :superadmin?
-  alias_method :state?,      :superadmin?
+  alias_method :supervisor?,       :superadmin?
+  alias_method :journals?,         :superadmin?
+  alias_method :comments?,         :superadmin?
+  alias_method :state?,            :superadmin?
+  alias_method :authority_fields?, :superadmin?
 end
--- a/app/policies/client_policy.rb
+++ b/app/policies/client_policy.rb
@@ -23,4 +23,5 @@ class ClientPolicy < ApplicationPolicy
   alias_method :supervisor?, :superadmin?
   alias_method :journals?,   :superadmin?
   alias_method :comments?,   :superadmin?
+  alias_method :state?,      :superadmin?
 end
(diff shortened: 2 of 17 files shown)


app/models/client.rb
```diff
@@ -72,4 +72,8 @@ def self.first_languages
   def self.state_collection
     STATES.map(&:to_sym)
   end
+
+  def german_missing?
+    language_skills.blank? || language_skills.german.blank?
```
chrysalag: Why is the first check needed?

app/policies/client_policy.rb
```diff
@@ -20,7 +20,9 @@ def resolve
   alias_method :need_accompanying?, :superadmin_or_department_manager?
 
   # suplementary policies
-  alias_method :supervisor?, :superadmin?
-  alias_method :journals?,   :superadmin?
-  alias_method :comments?,   :superadmin?
+  alias_method :supervisor?,       :superadmin?
+  alias_method :journals?,         :superadmin?
```
chrysalag: Since all those methods refer to the superadmin, why can't we have just one method?

test/system/clients_test.rb
```diff
@@ -94,30 +94,6 @@ class ClientsTest < ApplicationSystemTestCase
     end
   end
 
-  test 'language without a level shows only language' do
```
chrysalag: Maybe transfer and adapt the test to the `test/system/volunteers_test.rb`?

app/models/client.rb
casaper: Your right, its a relict from a version where there was also a default empty native_speker language added. But now this is obsolete. Going to remove it.

app/policies/client_policy.rb
casaper: to be able to control visibility explicitly in views, and also change those. Otherwise we end up asking `policy(Client).destroy?` for something in the views that has absoluteley nothing to do with destroy again.

chrysalag: I didn't suggest using `policy(Client).destroy?` to control view visibility. I suggested having one method eg `supervisor?` for all view visibility checks, since all of them check for the superadmin. After a while we won't remember which type of user should have access to those fields, unless we look up the client policy.

app/views/volunteers/_volunteer.html.slim
```diff
@@ -7,7 +7,7 @@ tr
   td = volunteer.contact.primary_email
   td = volunteer.birth_year.year if volunteer.birth_year?
   td = volunteer.native_language&.language_name
-  td = render 'language_skills/show_inline', language_skills: volunteer.foreign_languages
+  td = render 'language_skills/show_inline', language_skills: volunteer.language_skills.not_german
```
chrysalag: Isn't there any chance that a volunteer has another mother tongue than German?

app/policies/client_policy.rb
casaper: The approach would be that one would not worry about who can see what within a view. So within the view I would, if the Model-Field has any restrictions at all, ask policy if it should be displayed, and the granular tuning of who does what is done in the policy. If they are all the same now, it doesn't necessary stay like this forever. So if the view is built opon `policy(Client).state?` = Tell me if state should be displayed or not.

I don't state that the approach you suggest would be wrong. 
The approach I tried to implement was just to keep policy decisions as much as possible outside the views and within the policy. So It would be rather a decision if we pray in `app/views/clients/_form.html.slim` or in `app/policies/client_policy.rb`.
Dont argue about praying 😉

chrysalag: Then unless something needs to change regarding [this](https://github.com/panter/aoz-003/pull/315#discussion_r148503456), then is it ready to be merged?

casaper: mother tongue is beeing displayed in the colum before, and as you requested me to transfer the test for empty language levels in view displayed, it made this neeeded, because the scope used before was obviously not delivering `language without a level`. So Either you have your test or you have what ever you have!?!

chrysalag: I am only referring to the addition of `.not_german` here:

`td = render 'language_skills/show_inline', language_skills: volunteer.language_skills.not_german `

So the question was why:
`td = render 'language_skills/show_inline', language_skills: volunteer.language_skills.not_german`

and why not:

`td = render 'language_skills/show_inline', language_skills: volunteer.language_skills`

I don't see how that has anything to do with the test.

casaper: you know what. I'm dropping that test and put the index back to normal

PR #324, opened 2017-11-02
Changes: client edit show

chrysalag:
# [Story in Trello](https://trello.com/c/gB82fuDE/37-klienten-show-edit)

### Done for approval?

* [x] Story under test
* ~[ ] Mobile works~
* ~[ ] Seeds created~
* ~[ ] Translated~
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/views/clients/_columns.html.slim
+++ b/app/views/clients/_columns.html.slim
@@ -19,7 +19,8 @@
           span= sort_link @q, :competent_authority, t_attr(:competent_authority, Client)
         th.truncate-th
           span= sort_link @q, :created_at, t_attr(:created_at, Client)
-        th= t_attr(:comments, Client)
+        - if policy(Client).comments?
+          th= t_attr(:comments, Client)
         th.hidden-print colspan='2'
     tbody
       = render clients
--- a/app/views/clients/_columns.html.slim
+++ b/app/views/clients/_columns.html.slim
@@ -17,9 +17,9 @@
         th= t_attr(:interests, Client)
         th.truncate-th
           span= sort_link @q, :competent_authority, t_attr(:competent_authority, Client)
-        th= sort_link @q, :comments
         th.truncate-th
           span= sort_link @q, :created_at, t_attr(:created_at, Client)
+        th= t_attr(:comments, Client)
         th.hidden-print colspan='2'
     tbody
       = render clients
--- a/app/views/clients/_form.html.slim
+++ b/app/views/clients/_form.html.slim
@@ -1,22 +1,18 @@
 = simple_form_for(@client) do |f|
   = simple_error_notice f
-  = f.input :salutation, as: :select, collection: Client.salutation_collection,
-              include_blank: true
+  = f.input :salutation, as: :select, collection: Client::SALUTATIONS, include_blank: true
 
   .row
     = f.simple_fields_for :contact do |c|
-      .col-xs-12.col-md-6
-        = c.input :first_name, required: true
-      .col-xs-12.col-md-6
-        = c.input :last_name, required: true, label: t_attr(:last_name)
-    .col-xs-12.col-md-6
-      = f.input :birth_year, as: :select, collection: Client.year_collection, include_blank: true
-    .col-xs-12.col-md-6
-      = f.input :nationality, as: :country
+      .col-xs-12.col-md-6= c.input :first_name, required: true
+      .col-xs-12.col-md-6= c.input :last_name, required: true, label: t_attr(:last_name)
+    .col-xs-12.col-md-4= f.input :birth_year, as: :select, collection: Client.year_collection,
+                                 include_blank: true
+    .col-xs-12.col-md-4= f.input :entry_date
+    .col-xs-12.col-md-4= f.input :nationality, as: :country
 
   = f.simple_fields_for :contact do |c|
     = render 'contacts/fields', f: c
-
 
   fieldset
     legend= t_attr(:permit, Client)
@@ -37,7 +33,6 @@
       .links
         = link_to_add_polymorphic_association t('add_language'), f, :language_skills, class: 'btn btn-success'
 
-
   = render 'relatives/form', f: f, relative: @client
 
   fieldset
@@ -57,9 +52,9 @@
     fieldset
       legend Interne Kriterien
       = f.input :state, collection: Client.state_collection, include_blank: false
-      = f.input :comments
       = f.input :involved_authority
       = f.input :competent_authority
+      = f.input :comments
 
   fieldset
     legend= t('general_availability')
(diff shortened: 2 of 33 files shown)


app/views/clients/_columns.html.slim
```diff
@@ -15,8 +15,9 @@
         th= t_attr(:goals, Client)
         th= t_attr(:interests, Client)
         th= sort_link @q, :competent_authority, t_attr(:competent_authority, Client)
-        th= sort_link @q, :comments
         th= sort_link @q, :created_at, t_attr(:created_at, Client)
+        - if policy(Client).comments?
+          th= sort_link @q, :comments
```
nuritnt: I am not sure if sorting comments makes sense?

app/views/clients/_form.html.slim
```diff
@@ -1,52 +1,42 @@
...
-      = f.input :nationality, as: :country
-    .col-xs-12.col-md-6
-      = f.input :permit,
-                collection: Client::PERMITS,
-                as: :radio_buttons,
-                wrapper: :vertical_radio_and_checkboxes
+    .col-xs-12.col-md-4= f.input :nationality, as: :country
+    .col-xs-12.col-md-4= f.input :entry_date
+    .col-xs-12.col-md-4= f.input :permit, collection: Client::PERMITS, as: :radio_buttons,
+                                  wrapper: :vertical_radio_and_checkboxes
```
nuritnt: maybe change the checkboxes to horizontal and then add the entry date above the nationality?

app/views/clients/_show_navigation.html.slim
```diff
@@ -1,10 +1,9 @@
...
   ul.list-inline
-    li= button_link t_action(:edit), edit_client_path(@client)
-    - if policy(Journal).index?
-      li= button_link 'Journal', polymorphic_path([@client, Journal])
+    li= button_link t_action(:edit), edit_client_path(@client), 'success'
+    li= button_link t_action(:print), client_path(@client, print: true)
     - if policy(Client).destroy?
       li= link_to t_action(:destroy), @client, confirm_deleting(@client, 'btn btn-default')
+    - if policy(Journal).index?
+      li= button_link 'Journal', polymorphic_path([@client, Journal])
```
nuritnt: is it possible to align the journal button to the right? (since they asked for it to be on the right side)

chrysalag: Yes, I didn't realize exactly what they meant. Fixed.

app/views/clients/_form.html.slim
chrysalag: After the rebase the layout was different, so now again it is next to nationality, but now it looks good. What do you think?

app/views/clients/_show_navigation.html.slim
```diff
@@ -1,10 +1,10 @@
...
 nav.navbar.section-navigation
   ul.list-inline
-    li= button_link t_action(:edit), edit_client_path(@client)
-    - if policy(Journal).index?
-      li= button_link 'Journal', polymorphic_path([@client, Journal])
+    li= button_link t_action(:edit), edit_client_path(@client), 'success'
+    li= button_link t_action(:print), client_path(@client, print: true)
     - if policy(Client).destroy?
       li= link_to t_action(:destroy), @client, confirm_deleting(@client, 'btn btn-default')
     li.m-l-20= button_link t('back'), clients_path
```
nuritnt: I think this `m-l-20` is not needed anymore(there is a space before the back button)
screen shot 2017-11-03 at 18 25 05

app/views/clients/_form.html.slim
```diff
@@ -1,23 +1,19 @@
...
+    .col-xs-12.col-md-4= f.input :nationality, as: :country
+    .col-xs-12.col-md-4= f.input :entry_date
 
   = f.simple_fields_for :contact do |c|
     = render 'contacts/fields', f: c
 
-
   fieldset
     legend= t_attr(:permit, Client)
     = f.input :permit, collection: Client::PERMITS, as: :radio_buttons, label: false
```
nuritnt: I just meant to use `horizontal_radio_and_checkboxes` instead of the vertical ones. I think it should stay next to the nationality since it was requested back then. sorry for the misunderstanding..

app/views/clients/_show_navigation.html.slim
chrysalag: Changed

app/views/clients/_form.html.slim
chrysalag: adding `wrapper: :horizontal_radio_and_checkboxes` and unless it is part of the requested feedback, maybe it could stay as it is?

nuritnt: it was initially requested back then, when permit was implemented and I had a call with Zaida yesterday and she confirmed, that these 3 information should stay together

chrysalag: Do they have preference to radio buttons over dropdowns? Because I haven't yet found how to override the default settings and make the radio buttons horizontal.

db/migrate/20171103121704_add_entry_date_to_client.rb
```diff
@@ -0,0 +1,6 @@
+class AddEntryDateToClient < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :clients, :entry_year, :date
+    add_column :clients, :entry_date, :string
```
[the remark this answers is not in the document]
chrysalag: Probably yes, done. :)

db/migrate/20171103121704_add_entry_date_to_client.rb
```diff
@@ -1,6 +1,9 @@
 class AddEntryDateToClient < ActiveRecord::Migration[5.1]
   def change
-    remove_column :clients, :entry_year, :date
     add_column :clients, :entry_date, :string
+    Client.find_each do |client|
+      client.update(entry_date: client.entry_year)
+    end
```
toupeira: Note that this code will run both when running the migration and rolling it back, you can avoid that by defining `up` / `down` methods separately. 

Also I'm not sure if it will recognize the new `entry_date` column immediately, it used to be that you have to call `Client.reset_column_information` in the migration after adding new columns but I'm not sure if that's still the case in Rails 5. If the migration works locally it should be fine :)

chrysalag: It works locally. The thing is that in case of a rollback, how can the data be saved since the old column is `:date` and the new is `:string` and string cannot be stored in date?

app/views/clients/_form.html.slim
nuritnt: yes they requested radio buttons .. :/

chrysalag: Have they also requested them horizontal?

chrysalag: Ok, didn't realize it was optional. (It should definitely be possible, I just haven't found the way to do it.) So unless there aren't any other requested changes, it should be ready. :)

PR #330, opened 2017-11-03
Features/use chrome headless for system tests

casaper:
# [Story in Trello](#)

### Done for approval?

* [ ] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: ruby
 cache:
   bundler: true
   directories:
-    - .phantomjs
     - vendor
     - .rvm
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ addons:
 before_script:
   - bundle exec rails db:create
   - bundle exec rails db:schema:load
-  - export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start && sleep 3
 
 script:
   - bundle exec rails test
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: ruby
 cache:
   bundler: true
   directories:
+    - node_modules
     - vendor
     - .rvm
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,13 @@ cache:
 services: postgresql
 
 addons:
+  chrome: stable
   postgresql: "9.4"
 
 before_script:
   - bundle exec rails db:create
   - bundle exec rails db:schema:load
+  - export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start && sleep 3
 
 script:
   - bundle exec rails test
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -1,25 +1,35 @@
 require 'test_helper'
+require 'selenium/webdriver'
+
+Capybara.register_driver(:headless_chrome) do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: { args: %w[headless disable-gpu] }
+  )
+
+  Capybara::Selenium::Driver.new(
+    app,
+    browser: :chrome,
+    desired_capabilities: capabilities
+  )
+end
+
+Capybara.register_driver(:local_headless_chrome) do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: { args: %w[headless] }
+  )
+
+  Capybara::Selenium::Driver.new(
+    app,
+    browser: :chrome,
+    desired_capabilities: capabilities
+  )
+end
+
+Capybara.register_driver(:visible_chrome) do |app|
+  Capybara::Selenium::Driver.new(app, browser: :chrome)
+end
 
 DIMENSIONS = [1800, 4000].freeze
-
-Capybara.register_driver :poltergeist_debug do |app|
-  Capybara::Poltergeist::Driver.new(
-    app,
-    phantomjs: Phantomjs.path,
-    inspector: true,
-    screen_size: DIMENSIONS,
-    window_size: DIMENSIONS
-  )
-end
-
-Capybara.register_driver :poltergeist do |app|
-  Capybara::Poltergeist::Driver.new(
-    app,
-    phantomjs: Phantomjs.path,
-    screen_size: DIMENSIONS,
-    window_size: DIMENSIONS
-  )
-end
 
 def driven_by_default(driver = :poltergeist, using: nil)
   driven_by driver, using: using, screen_size: DIMENSIONS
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -3,18 +3,20 @@ language: ruby
...
 services: postgresql
 
 addons:
+  chrome: stable
   postgresql: "9.4"
 
 before_script:
   - bundle exec rails db:create
   - bundle exec rails db:schema:load
+  - export DISPLAY=:99.0 && sh -e /etc/init.d/xvfb start && sleep 3
```
toupeira: You can also try using `xvfb-run` instead, i.e. `xvfb-run bundle exec rails test` in the `script:` section below. This starts Xvfb, sets `$DISPLAY`, runs the command when Xvfb is ready, and kills Xvfb again when the command has finished.

toupeira: Hmm actually, do we really need Xvfb? I haven't used Chrome Headless yet, but everything I can find about it so far seems to indicate that it's indeed truly headless, i.e. it shouldn't need an X server at all.

test/application_system_test_case.rb
```diff
@@ -1,38 +1,42 @@
...
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   case ENV['driver']
-  when 'chrome'
-    driven_by_default :selenium, using: :chrome
-  when 'poltergeist_debug'
-    driven_by_default :poltergeist_debug
+  when 'visible'
+    driven_by :visible_chrome
+  when 'faster' # maybe, don't know if gpu helps at all
+    driven_by :local_headless_chrome
```
toupeira: GPU definitely shouldn't hurt, does it cause problems on Travis? If not I would just keep it enabled and combine `headless_chrome` / `local_headless_chrome`

test/integration/clients_xlsx_export_test.rb
```diff
@@ -15,12 +15,11 @@ def setup
   end
 
   test 'xlsx files columns and cells are correct' do
-    client_older = create :client
     client = create :client, entry_year: 2.years.ago, birth_year: 30.years.ago,
       education: 'educati', created_at: 2.days.ago
 
     # order id desc to be sure to have the right record at row 2
-    get clients_url(format: :xlsx, q: { s: 'id+desc' })
+    get clients_url(format: :xlsx)
```
toupeira: The comment above still references the query parameters ;-)

test/integration/clients_xlsx_export_test.rb
```diff
@@ -63,8 +62,6 @@ def setup
     assert_equal I18n.t("state.#{client.state}"),            wb.cell(2, 16)
     assert_equal client.created_at.to_date,                  wb.cell(2, 17).to_date
     assert_equal client.updated_at.to_date,                  wb.cell(2, 18).to_date
-
-    assert_equal client_older.id,                            wb.cell(3, 1)
```
toupeira: Why was this removed?

test/system/clients_test.rb
```diff
@@ -118,11 +118,14 @@ class ClientsTest < ApplicationSystemTestCase
     refute page.has_text? 'Fluent'
   end
 
-  test 'superadmin can delete client' do
+  test 'superadmin_can_delete_client' do
```
toupeira: Accidental change? :)

.travis.yml
casaper: I've simply taken examples from the net. They where using it, but of course I can try and see if it runs without

test/application_system_test_case.rb
casaper: but what will travis do, with gpu support? I don't know. This one I've taken over from examples as well, without actually knowing if it's correct.
I basicly put this there to try out if there is any difference. But if travis won't complain about gpu_support, I don't see a reason to have it disabled at all. So yea, would maybe make sense.

test/integration/clients_xlsx_export_test.rb
casaper: because the test kept on failing sometimes. I don't know why Id's and order where not predictable in tests. In manual testing it always worked perfectly fine.

test/system/clients_test.rb
casaper: nope, not really. Snake case test names are easier to call as single run by `rails test -n /superadmin_can_delete_client/ test/system/clients_test.rb`. This way it can be copy and pasted. Doing `rails test -n '/superadmin can delete client/' test/system/clients_test.rb` did not work here usually.
For unknown reason I was not able to run our system tests by line as documented by rails.

casaper: nope, not really. Snake case test names are easier to call as single run by rails test -n /superadmin_can_delete_client/ test/system/clients_test.rb. This way it can be copy and pasted. Doing rails test -n '/superadmin can delete client/' test/system/clients_test.rb did not work here usually.
For unknown reason I was not able to run our system tests by line as documented by rails.

casaper: Maybe you know why I canot run `rails test test/system/clients_test.rb:122`.
Here I still get all the tests in the file executed, if I do that

.travis.yml
casaper: I guess in some or another form xvfb is needed. I'm just guessing: Maybe it is needed for the screenshots?

For example FutureKids uses chrome headless since ages, and there xvfb is also run: https://github.com/panterch/future_kids/blob/master/.travis.yml

test/system/clients_test.rb
chrysalag: This test now starts at line 121 not 122. (Although locally both commands run successfully. )

casaper:
@chrysalag firefox is not used here. It's all done with chrome.

test/system/clients_test.rb
casaper: that should not matter. It runs only one single test from the file on your box?

.travis.yml
toupeira: There's some confusion with the terms:

- `chromedriver` is a version of Chrome that can be remotely controlled through the WebDriver API (using Selenium or others) and requires Xvfb (this is what FutureKids uses)
- `chrome --headless` is a new feature in Chrome 59 which supports running Chrome without Xvfb (this is what we're now using)

I tried to find out more and found this on https://developers.google.com/web/updates/2017/04/headless-chrome:

> **So I still need Xvfb?**
>
> No. Headless Chrome doesn't use a window so a display server like Xvfb is no longer needed. You can happily run your automated tests without it.
>
> What is Xvfb? Xvfb is an in-memory display server for Unix-like systems that enables you to run graphical applications (like Chrome) without an attached physical display. Many people use Xvfb to run earlier versions of Chrome to do "headless" testing.

Also noticed this question above:

> **Do I need the --disable-gpu flag?**
>
> Yes, for now. The --disable-gpu flag is a temporary requirement to work around a few bugs. You won't need this flag in future versions of Chrome. See https://crbug.com/546953#c152 and https://crbug.com/695212 for more information.

So I guess that's still needed for now :)

test/integration/clients_xlsx_export_test.rb
toupeira: Yeah I was wondering if it's related to the removal of the sort parameter above :)

```ruby
# order id desc to be sure to have the right record at row 2
get clients_url(format: :xlsx, q: { s: 'id+desc' })
```

test/system/clients_test.rb
toupeira: Hmm both commands run all tests as well here, even though [passing a line number seems to be supported](http://guides.rubyonrails.org/testing.html#the-rails-test-runner)

Same with model tests, so it's not related to system tests either.

PR #338, opened 2017-11-08
WIP: Features/fuzzy client search

nuritnt:
# Had to be reverted and reopened, since all names where not shown anymore on the indexes on staging

# [Story in Trello](https://trello.com/c/Uz0BSVr5/42-suchfunktion)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [ ] Seeds created
* [x] Code is reviewed

### Done after the merge?


nuritnt:
@toupeira Do you maybe have an idea why all names on the indexes disappeared when deploying on staging?

test/application_system_test_case.rb
```diff
@@ -51,4 +51,26 @@ def select_date(selects, *values)
     selects[1].select values[1]
     selects[2].select values[2]
   end
+
+  def wait_for_ajax
+    Timeout.timeout(Capybara.default_max_wait_time) do
+      active = page.evaluate_script('jQuery.active')
+      active = page.evaluate_script('jQuery.active') until active.zero?
+    end
```
toupeira: Haha [this sounds familiar](https://github.com/panter/gsc/blob/d4419296c82bab96de71d428efe03ccf268d261a/spec/support/capybara_helper.rb#L123-L127) ;-)

You can simplify it to `loop until page.evaluate_script('jQuery.active').zero?`

test/application_system_test_case.rb
```diff
@@ -51,4 +51,26 @@ def select_date(selects, *values)
...
+    items = page.find_all('li.ui-menu-item')
+    if options[:items_expected]
+      assert_equal options[:items_expected], items.size
+    end
+    if options[:check_items].present?
+      items.each do |item|
+        assert options[:check_items].include? item.text
+      end
+    end
+    find("[name=\"#{name}\"]").native.send_keys :down, :enter
```
toupeira: Maybe do `input = find("[name=\"#{name}\"]")` at the start of the method

app/controllers/clients_controller.rb
```diff
@@ -18,6 +18,15 @@ def index
     end
   end
 
+  def search
+    authorize Client
+    @q = policy_scope(Client).ransack contact_full_name_cont: params[:term]
```
toupeira: You could also do something like `contact_first_name_or_contact_last_name_cont` to avoid having to cache the `full_name` value in the DB.

Ransack also supports [defining aliases](https://github.com/activerecord-hackery/ransack#ransack-aliases) as a shortcut.

toupeira: I guess the downside is it won't support searching for the full name directly, but not sure if the users will expect to be able to do that (they will have to include the comma too)

casaper:
@nuritnt & @toupeira 
> Do you maybe have an idea why all names on the indexes disappeared when deploying on staging?

I do. I simply forgot about the allready existing records. I made a migration now, that updates there newly added full_name fields from their last and first name fields.

test/integration/clients_search_suggestion_test.rb
```diff
@@ -0,0 +1,22 @@
...
+      [client_one, client_two]
+    end
+    login_as superadmin
+    get search_clients_path, as: :json, params: { term: 'aaa' }
+    results = JSON.parse response.body
+    assert_equal 2, results.size
+    results.map { |result| result['value'] }.each do |result|
+      assert clients.first.map { |client| client.contact.full_name }.include? result
+      refute clients.last.map { |client| client.contact.full_name }.include? result
+    end
```
toupeira: This test is a bit hard to read, why not simply check the two result values directly?

PR #343, opened 2017-11-09
Features/Erinnerungslauf initieren: Probezeit

chrysalag:
# [Story in Trello](https://trello.com/c/Oeox3mhz/6-erinnerungslauf-initieren-probezeit)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* ~[ ] Seeds created~
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/controllers/reminders_controller.rb
+++ b/app/controllers/reminders_controller.rb
@@ -3,7 +3,7 @@ class RemindersController < ApplicationController
 
   def index
     authorize Reminder
-    @reminders = Reminder.all
+    @reminders = Reminder.assignment
   end
 
   # Sends the email
@@ -17,6 +17,12 @@ class RemindersController < ApplicationController
   def destroy
     @reminder.destroy
     redirect_to reminders_url, make_notice
+  end
+
+  def trial_end
+    authorize Reminder
+    Reminder.trial_end_reminders
+    @trial_end = Reminder.trial
   end
 
   private
--- a/app/controllers/reminders_controller.rb
+++ b/app/controllers/reminders_controller.rb
@@ -16,6 +16,7 @@ class RemindersController < ApplicationController
 
   def destroy
     @reminder.destroy
+    @reminder.volunteer.update(trial_period: true) if @reminder.trial?
     redirect_to @reminder.trial? ? trial_end_reminders_url : reminders_url, make_notice
   end
 
--- a/app/controllers/reminders_controller.rb
+++ b/app/controllers/reminders_controller.rb
@@ -16,7 +16,7 @@ class RemindersController < ApplicationController
 
   def destroy
     @reminder.destroy
-    redirect_to reminders_url, make_notice
+    redirect_to @reminder.trial? ? trial_end_reminders_url : reminders_url, make_notice
   end
 
   def trial_end
--- a/app/models/concerns/group_assignment_and_assignment_scopes.rb
+++ b/app/models/concerns/group_assignment_and_assignment_scopes.rb
@@ -19,7 +19,9 @@ module GroupAssignmentAndAssignmentScopes
     scope :start_after, ->(date) { where('period_start > ?', date) }
     scope :start_within, ->(date_range) { where(period_start: date_range) }
     scope :started_six_months_ago, (-> { where('period_start < ?', 6.months.ago.to_date) })
-    scope :started_six_weeks_ago, (-> { where('period_start < ?', 6.weeks.ago.to_date) })
+    scope :started_ca_six_weeks_ago, lambda {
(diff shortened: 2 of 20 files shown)


app/models/reminder.rb
```diff
@@ -50,6 +58,11 @@ def self.conditionally_create_reminder_for_volunteer(volunteer, assignment)
     end
   end
 
+  def self.conditionally_create_trial_end_reminder_for_volunteer(volunteer, a)
+    Reminder.create!(volunteer: volunteer, assignment: a, kind: 0) if a.started_ca_six_weeks_ago? &&
```
nuritnt: would prefer to use `assignment` instead of only using `a`. didn't abbreviated for the methods above, and its clearer.

app/controllers/reminders_controller.rb
```diff
@@ -3,7 +3,7 @@ class RemindersController < ApplicationController
 
   def index
     authorize Reminder
-    @reminders = Reminder.all
+    @reminders = Reminder.assignment
```
nuritnt: since we don't have the `Reminder.all` here, what will be shown at the `"normal" Reminders Index` ?

app/models/concerns/group_assignment_and_assignment_scopes.rb
```diff
@@ -18,7 +18,10 @@ module GroupAssignmentAndAssignmentScopes
     scope :start_before, ->(date) { where('period_start < ?', date) }
     scope :start_after, ->(date) { where('period_start > ?', date) }
     scope :start_within, ->(date_range) { where(period_start: date_range) }
-    scope :started_six_months_ago, (-> { where('period_start < ?', 6.months.ago.to_date) })
+    scope :started_six_months_ago, (-> { where('period_start < ?', 6.months.ago) })
+    scope :started_ca_six_weeks_ago, lambda {
+      where('period_start < ? AND period_start > ?', 6.weeks.ago, 8.weeks.ago)
```
nuritnt: I saw that in the Story it is written `after "ca" 6 Wochen` but what kind of difference does it make with these two time frames? Can't we just simplify it, and say after 6 Weeks? It will show up anyway after 6 weeks right? So why is there the second date needed?

chrysalag: After 6 weeks unlimited will show all (group) assignments that 6 weeks or older. So (group) from the beginning of time until 6 weeks ago.

app/models/reminder.rb
chrysalag: Me too, I did that because RuboCop complains about the line length.

app/controllers/reminders_controller.rb
chrysalag: The 'normal' index now will show reminders that concern assignments excluding the ones referring to the trial period. When the assignment reminders are produced through the cron job `kind` is set to `assignment`, when the trial period are produced `kind` is set to `trial`. So  `Reminder.assignment` and `Reminder.trial` work similarly to scopes.

app/models/concerns/group_assignment_and_assignment_scopes.rb
nuritnt: ahso! hmmm so after the 8 weeks, this won't show up in the reminders list huh? 
If we are sure, that they always get feedback within this two weeks frame, then I agree with this. Else, they will loose the overview from the missing feedback, which wasn't given in this two weeks frame..

app/controllers/reminders_controller.rb
nuritnt: 👍

app/models/concerns/group_assignment_and_assignment_scopes.rb
chrysalag: We are not sure. But otherwise it would show assignments for which they don't want/need to send reminders and they would have to delete them in order not to show up again. The time frame and the reminders are not associated with the feedbacks yet, because they don't exist yet. Otherwise we could set as scope at the 6 weeks and modify the reminder criteria to take into account the existence of feedback.

app/views/feedbacks/_main_fields.html.slim
```diff
@@ -1,9 +1,6 @@
-
 - [:goals, :achievements, :future, :comments, :conversation].each do |field|
   .row
-    .col-xs-12
-      = f.input field
+    .col-xs-12= f.input field
```
nuritnt: :field ?

nuritnt: ah no, not needed. nevermind

chrysalag: This is part of a loop and `field` is a variable.

PR #348, opened 2017-11-10
Refactor volunteer state implementation

casaper:
# [Story in Trello](https://trello.com/c/C7VW409p/16-reimplementation-of-volunteer-state-active-or-inactive)

### Done for approval?

* [x] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [x] Code is reviewed

### Done after the merge?


app/models/assignment.rb
```diff
@@ -26,7 +27,21 @@ class Assignment < ApplicationRecord
   scope :created_before, ->(max_time) { where('assignments.created_at < ?', max_time) }
   scope :created_after, ->(min_time) { where('assignments.created_at > ?', min_time) }
 
-  scope :ended, (-> { where('assignments.period_end < ?', Time.zone.today) })
+  scope :active, lambda {
```
chrysalag: Currently we have those concerns defined in a concern. If they are to be moved in the models again, then they should be deleted from the concern.

app/models/volunteer.rb
```diff
@@ -113,35 +113,57 @@ class Volunteer < ApplicationRecord
...
+    end
+  end
+
+  def period_end_maxes
+    [group_assignments.active.maximum(:period_end), assignments.active.maximum(:period_end)]
+  end
+
   def active?
-    accepted? && (assignments.active.any? || group_assignments.ongoing.any? || group_assignments.no_end.any?)
+    accepted? && assignments.active.any? || group_assignments.active.any?
```
chrysalag: Parenthesis is needed because `&&` has higher [priority](https://ruby-doc.org/core-2.2.0/doc/syntax/precedence_rdoc.html) than `||`.

app/models/volunteer.rb
```diff
@@ -113,35 +113,57 @@ class Volunteer < ApplicationRecord
...
-      .or(loj_without_assignments)
-      .or(loj_active_take_more)
+    seeking_clients.or(accepted.will_take_more_assignments)
   }
 
+  def verify_and_update_state
+    update(active: active?, activeness_might_end: relevant_period_end_max)
+  end
+
+  def relevant_period_end_max
```
chrysalag: There is some implementation of the max dates from the new line 186 of this used in certificates for the computation of duration start and end. If the old implementation is to change then the lines 186-... are not needed and then the duration of the certificates has to be adjusted.

app/controllers/volunteers_controller.rb
```diff
@@ -66,6 +68,15 @@ def seeking_clients
 
   private
 
+  def activeness_filter
+    return unless params[:q] && params[:q][:active_eq]
+    if params[:q][:active_eq] == 'true'
```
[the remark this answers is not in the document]
casaper: no, params 'true' is no boolean. Its a string

app/helpers/filter_dropdown_helper.rb
```diff
@@ -24,6 +24,34 @@ def enum_filter_dropdown(attribute, collection)
...
+  def activeness_filter_dropdown
+    li_dropdown do
+      concat dropdown_toggle_link('Status' + status_toggler_text_end)
+      concat dropdown_ul(tag.li { all_link_to(:active_eq) }) { activeness_links }
+    end
+  end
+
+  def status_toggler_text_end
+    return ' ' unless search_parameters['active_eq']
+    if search_parameters['active_eq'] == 'true'
```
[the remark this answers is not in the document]
casaper: no, its not a boolean

app/models/assignment.rb
casaper: they have `assignments.created_at` in where, so will only work on Assignment, and no where else

app/controllers/volunteers_controller.rb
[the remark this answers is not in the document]
chrysalag: Sorry, I didn't notice the quotes.

app/models/volunteer.rb
casaper: Cannot be reused here, because this implementations requires knowledge of nil, and the one you mention filters nil cases

app/models/volunteer.rb
casaper: nope, your wrong. This is boolean...
```ruby
true && false || true  # returns true
true && true || false # returns true
true && true || true # returns true
true && false || false # returns false
```
Try it in your shell if you don't beleave me. This is simply how boolean logic works

casaper: paranteses makes no difference, they are redundant. Boolean logic in Ruby works from left to right

chrysalag: If we need the check to be done from left to right, the parenthesis not only is not needed, but also would be wrong to use. 
The reason that this is executed from left to right is that logic-boolean and has higher priority than logic-boolean or. That is not my personal opinion. It is suggested in the link provided and it is a rule in all logic operations. 
The reason is not that they have have the same priority. If they did, then, of course, again it would be executed from left to right. Operation of equal priority are indeed always executed from left to right.

chrysalag: In any case `false && false || true ` will give true and I guess we wouldn't like a volunteer that is not accepted to be considered as active.

casaper: damn, I hate to be wrong 🤦‍♂️ ... your right

app/models/volunteer.rb
casaper: Ah, I see now what you mean. I was wrong again. I can use those methods, but as looking at them, I found that they had some flaw, because they where not useing `.pluck(:period_end)`.
So after fixing them, I will make use of these here. Thx

test/models/volunteer_state_test.rb
```diff
@@ -0,0 +1,209 @@
+require 'test_helper'
+
+class VolunteerStateTest < ActiveSupport::TestCase
+  def create_basic_accepted_volunteer
+    volunteer_user = create :user_volunteer, volunteer: create(:volunteer)
+    volunteer_user.volunteer
```
chrysalag: What is this doing?

app/models/assignment.rb
```diff
@@ -26,7 +27,21 @@ class Assignment < ApplicationRecord
...
+      Time.zone.today
+    )
+  }
+  scope :started, lambda {
+    where(
+      'assignments.period_start < ? AND assignments.period_start IS NOT NULL',
+      Time.zone.today
+    )
+  }
+  scope :ended, lambda {
```
chrysalag: Also, out of curiosity, what is the advantage of using `lambda`?

test/models/volunteer_state_test.rb
casaper: returns the `Volunteer`, not the `User`

app/models/assignment.rb
casaper: rubocop does not accept multiline lambdas like this:
```ruby
scope :example, (-> { where(something: 'xy') })  # is rubocophappy
scope :example, (-> {  # alltho this would AFAIR work, rubocop hates it
  where(x: 'd')
    ........
})
```

So thats why.
In general i dont like the syntax of those lambas anyway, because these three are all technicly lambda, but are syntactically inconsistent:
```ruby
scope :rara, (-> { xy })
scope :rara, (param)-> { where(x: param) }
scope :rara, lambda {
   ....
}
scope :rara, lambda { |param|
   ....
}
```

🤢

test/models/volunteer_state_test.rb
casaper: @chrysalag but as you mention it, actually it isn't so nice. I made a sub factory of volunteer `:volunteer_with_user` that creates a volunteer, that has a volunteer role user with it.

PR #349, opened 2017-11-14
Generate Excel list for local MS-Office Mass Mailings

casaper:
# [Story in Trello](https://trello.com/c/VM4RXFXu/1-excel-listen-mit-emails-generieren)

### Done for approval?

> Except this one in question: "FW-Verantwortliche sollen nur ihre eigenen FW anschreiben können"

* [x] Story under test
* [ ] Mobile works
* [ ] Seeds created
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/helpers/filter_dropdown_helper.rb
+++ b/app/helpers/filter_dropdown_helper.rb
@@ -41,30 +41,32 @@ module FilterDropdownHelper
   def external_internal_links
     params_u = params.to_unsafe_hash.except('page')
     [
-      tag.li do
+      concat(tag.li do
         link_to('Extern ', url_for(params_u.merge(q: search_parameters.merge(external_eq: 'true'))))
-      end,
-      tag.li do
+      end),
+      concat(tag.li do
         link_to('Intern ', url_for(params_u.merge(q: search_parameters.merge(external_eq: 'false'))))
-      end
-    ].collect { |li| concat li }
+      end)
+    ].collect
   end
 
   def external_toggler_text_end
-    return ' ' unless search_parameters['active_eq']
     if search_parameters['external_eq'] == 'true'
       ': Extern '
     elsif search_parameters['external_eq'] == 'false'
       ': Intern '
+    else
+      ' '
     end
   end
 
   def status_toggler_text_end
-    return ' ' unless search_parameters['active_eq']
     if search_parameters['active_eq'] == 'true'
       ': Aktiv '
     elsif search_parameters['active_eq'] == 'false'
       ': Inaktiv '
+    else
+      ' '
     end
   end
 
--- a/app/helpers/filter_dropdown_helper.rb
+++ b/app/helpers/filter_dropdown_helper.rb
@@ -31,6 +31,34 @@ module FilterDropdownHelper
     end
   end
 
+  def intern_extern_filter_dropdown
+    li_dropdown do
+      concat dropdown_toggle_link('Intern/Extern' + external_toggler_text_end)
+      concat dropdown_ul(tag.li { all_link_to(:external_eq) }) { external_internal_links }
+    end
+  end
+
+  def external_internal_links
+    params_u = params.to_unsafe_hash.except('page')
+    [
+      tag.li do
+        link_to('Extern ', url_for(params_u.merge(q: search_parameters.merge(external_eq: 'true'))))
+      end,
+      tag.li do
+        link_to('Intern ', url_for(params_u.merge(q: search_parameters.merge(external_eq: 'false'))))
+      end
+    ].collect { |li| concat li }
+  end
+
+  def external_toggler_text_end
+    return ' ' unless search_parameters['active_eq']
+    if search_parameters['external_eq'] == 'true'
+      ': Extern '
+    elsif search_parameters['external_eq'] == 'false'
+      ': Intern '
+    end
+  end
+
   def status_toggler_text_end
     return ' ' unless search_parameters['active_eq']
     if search_parameters['active_eq'] == 'true'
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,7 +14,7 @@ class User < ApplicationRecord
   has_many :journals
   has_many :assignments, inverse_of: 'creator'
   has_many :feedbacks, inverse_of: 'author'
-  has_many :billing_expenses
+  has_many :billing_expense
   has_many :group_offers, inverse_of: 'creator'
 
   has_and_belongs_to_many :department
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,7 +14,7 @@ class User < ApplicationRecord
   has_many :journals
   has_many :assignments, inverse_of: 'creator'
   has_many :feedbacks, inverse_of: 'author'
-  has_many :billing_expense
+  has_many :billing_expenses
   has_many :group_offers, inverse_of: 'creator'
 
   has_and_belongs_to_many :department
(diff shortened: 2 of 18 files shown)


app/helpers/filter_dropdown_helper.rb
```diff
@@ -31,6 +31,34 @@ def activeness_filter_dropdown
...
+  def external_internal_links
+    params_u = params.to_unsafe_hash.except('page')
+    [
+      tag.li do
+        link_to('Extern ', url_for(params_u.merge(q: search_parameters.merge(external_eq: 'true'))))
+      end,
+      tag.li do
+        link_to('Intern ', url_for(params_u.merge(q: search_parameters.merge(external_eq: 'false'))))
+      end
+    ].collect { |li| concat li }
```
toupeira: Alternatively, you can also call `concat` directly before each tag (i.e. `concat tag.li do ...`)

app/models/user.rb
```diff
@@ -14,7 +14,7 @@ class User < ApplicationRecord
   has_many :journals
   has_many :assignments, inverse_of: 'creator'
   has_many :feedbacks, inverse_of: 'author'
-  has_many :billing_expenses
+  has_many :billing_expense
```
toupeira: Why not leave it as plural? ;-)

app/models/volunteer.rb
```diff
@@ -286,6 +287,16 @@ def to_s
 
   private
 
+  def user_added?
+    saved_change_to_attribute?(:user_id)
+  end
+
+  def copy_contact_to_user
+    user.profile&.contact&.update(contact.slice(:first_name, :last_name, :street, :postal_code,
+      :city, :primary_phone, :secondary_phone, :primary_email))
```
toupeira: Is it okay to always overwrite the existing contact, or do we need to preserve existing values?

app/views/users/index.xlsx.axlsx
```diff
@@ -0,0 +1,69 @@
+wb = xlsx_package.workbook
+col_header = wb.styles.add_style(
+    bg_color: 'FFDFDEDF',
+    b: true,
+    alignment: { horizontal: :center, vertical: :center },
+    border: { color: '00', edges: [:bottom], style: :thin}
+)
+standard_format = wb.styles.add_style alignment: { horizontal: :left }, width: :auto_fit
+date_time_format = wb.styles.add_style format_code: 'dd.mm.yyyy HH:MM'
+date_format = wb.styles.add_style format_code: 'dd.mm.yyyy'
```
toupeira: We already have these in `app/views/assignments/_assignments_table.xlsx.axlsx` and `app/views/clients/_client_table.xlsx.axlsx`, maybe time to extract them into a helper? e.g. a method `excel_styles` which returns a memoized hash of the styles (`@excel_styles ||= { ... }`)

app/views/volunteers/index.html.slim
```diff
@@ -13,10 +13,11 @@ nav.navbar.section-navigation
   ul.list-inline
     li= button_link t('clear_filters'), volunteers_path, 'info', dimension: 'sm'
     li= button_link t('export_xlsx'), url_for(format: :xlsx, q: search_parameters), 'warning', dimension: 'sm'
-    - unless params[:q] && params[:q][:acceptance_eq]
+    - unless search_parameters.keys.include? 'acceptance_eq'
       = activeness_filter_dropdown
-    - unless params[:q] && params[:q][:status_eq]
+    - unless search_parameters.keys.include? 'active_eq'
```
toupeira: On hashes `.include?` already does the same as `.keys.include?` ;-)

test/factories/volunteers.rb
```diff
@@ -72,8 +72,7 @@
 
     factory :volunteer_with_user do
       after(:create) do |volunteer|
-        volunteer.user = FactoryBot.create(:user_volunteer)
-        volunteer.save
+        volunteer.update(user: FactoryBot.create(:user_volunteer))
```
toupeira: You can just call `create` directly, no need for `FactoryBot.`

test/integration/users_xlsx_export_test.rb
```diff
@@ -0,0 +1,60 @@
...
+    contact = subject_user.profile.contact
+    assert_equal subject_user.id.to_s, wb.cell(row, 1).to_s
+    assert_xls_cols_equal(wb, row, 1, contact.full_name, contact.street, contact.extended,
+      contact.postal_code,contact.city, contact.primary_phone)
+    assert_xls_cols_equal(wb, row, 7, subject_user.email)
+  end
+
+  test 'xlsx file is downloadable' do
+    get users_url(format: :xlsx)
+    assert_equal('application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
```
toupeira: `assert_equal(Mime[:xlsx], response.content_type)` ;-)

app/models/volunteer.rb
casaper: I hit into the problem that exporting users with their `user.profile.contact` data gave me trouble on the volunteer users. Differing names etc., so thats why i did this. Assuming the very most likely case, that a volunteer gets associated with a user only once ever, this is only run once ever.

It was the solution that came up for me, but maybe not perfect. On the other hand it would be also quite miserable to check the `user.role` when exporting the xlsx.

I like neither of the solutions is nice. This is due to the fact that Volunteer has two contact records. One directly and anotherone after Volunteer.user.profile.contact has been created.

app/views/users/index.xlsx.axlsx
casaper: I think those styles need to be passed to the actual WorkBook object, so this would only be possible via passing `wb` to the helper.
Using a pratial could be the other option to do it.
But is that realy obvious and explicit?

test/factories/volunteers.rb
casaper: I did it because I assumed create to be in the actual objects scope somehow.... very most likely you're right 😉

app/models/volunteer.rb
toupeira: Okay in that case just leave it as it is now.

Not sure what a clean solution would look like, it would probably make sense to use the contact directly on the volunteer only until the user is created, and then remove the volunteer's contact and only use the user's profile contact from then on.

app/views/users/index.xlsx.axlsx
toupeira: Ah right missed that, it does look a bit ugly if we have to pass the workbook :)

What we could do instead is have a helper to create a new workbook with our default styles, i.e.:

```ruby
module ExcelHelper
  def workbook_with_styles
    wb = xlsx_package.workbook
    wb.styles.add_style[...]
    wb
  end
end

# and then in the view
wb = workbook_with_styles
```

test/factories/volunteers.rb
toupeira: We have `include FactoryBot::Syntax::Methods` in `test/test_helper.rb` so the methods are available everywhere in test cases :)

app/views/users/index.xlsx.axlsx
casaper: that looks convincing to me. I'll do it then.

casaper: nein, leider fehlanzeige, weil xlsx_package irgend wie nur in views zur verfügung steht:
https://github.com/straydogstudio/axlsx_rails/blob/master/lib/axlsx_rails/template_handler.rb

toupeira: Hmm strange, the helper runs in the same scope as the view so this shouldn't matter. But you're right, it's not really worth the effort ;-)

PR #351, opened 2017-11-14
Features/confirmation on client registration

nuritnt:
# [Story in Trello](https://trello.com/c/eBObdDXt/3-bei-der-anmeldung-des-kl-sozialarbeiter-in-text-anzeigen)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Code is reviewed

### Done after the merge?


app/views/client_notifications/new.html.slim
```diff
@@ -0,0 +1,5 @@
+h1= 'Klienten Wartezeit Benachrichtigung'
```
casaper:
```slim
h1 Klienten ..b.... blabla
```

app/controllers/client_notifications_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  private
+
+  def set_client_notification
+    @client_notification = ClientNotification.find(params[:id])
+    authorize @client_notification
+  end
+
+  def translate_model_name
+    @translated_model_name = @client_notification.class.model_name.human
```
casaper:
```ruby
def translate_model_name
  @client_notification.class.model_name.human
end
```

app/controllers/client_notifications_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def edit; end
+
+  def create
+    @client_notification = ClientNotification.new(client_notification_params)
+    @client_notification.user = current_user
+    authorize @client_notification
+    if @client_notification.save
+      redirect_to client_notifications_path,
+        notice: t('crud.created', model: @client_notification.class.model_name.human)
```
casaper: change the method `translate_model_name` and use it here:

```ruby
redirect_to client_notifications_path, notice: t('crud.created', model: translate_model_name)
```

app/controllers/client_notifications_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    else
+      render :new
+    end
+  end
+
+  def update
+    if @client_notification.update(client_notification_params)
+      redirect_to client_notifications_path, notice: t('crud.updated', model: @translated_model_name)
```
casaper: change the method `translate_model_name` and use it here:

```ruby
redirect_to client_notifications_path, notice: t('crud.created', model: translate_model_name)
```

app/controllers/client_notifications_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    else
+      render :edit
+    end
+  end
+
+  def destroy
+    @client_notification.destroy
+    redirect_to client_notifications_url, notice: t('crud.destroyed', model: @translated_model_name)
```
casaper: change the method `translate_model_name` and use it here:

```ruby
redirect_to client_notifications_path, notice: t('crud.created', model: translate_model_name)
```

app/controllers/client_notifications_controller.rb
```diff
@@ -0,0 +1,56 @@
+class ClientNotificationsController < ApplicationController
+  before_action :set_client_notification, only: [:edit, :update, :destroy]
+  before_action :translate_model_name, only: [:update, :destroy]
```
casaper: after the translate model name method is changed like explained below, you can drop this before_action

db/migrate/20171116140737_create_client_notifications.rb
```diff
@@ -0,0 +1,12 @@
+class CreateClientNotifications < ActiveRecord::Migration[5.1]
+  def change
+    create_table :client_notifications do |t|
+      t.text :body
+      t.references :user, foreign_key: true
+      t.boolean :active
+      t.datetime :deleted_at
```
casaper: Just for your information, no need to change this:
```ruby
t.datetime :deleted_at, index: true
```

this way the add_index at the bottom is not needed.

app/controllers/client_notifications_controller.rb
casaper: PS: Maybe rename `translate_model_name` to something in the manner of `human_model_name` or something else you find explicit.

app/views/clients/_listing_menu.html.slim
```diff
@@ -18,6 +18,8 @@ nav.navbar.section-navigation.hidden-print
       li= button_link 'Klienten ohne Begleitung', need_accompanying_clients_url, dimension: :sm
     - if action_name == 'need_accompanying'
       li= button_link 'Klienten mit Begleitung', with_assignment_clients_url, dimension: :sm
+    - if policy(ClientNotification).clients_index?
+      li= button_link 'Wartezeit Benachrichtigungen', client_notifications_path, dimension: 'xl'
```
chrysalag: Why `dimension: 'xl'`?
(Also there is an extra empty line after that.)

app/controllers/client_notifications_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def edit; end
+
+  def create
+    @client_notification = ClientNotification.new(client_notification_params)
+    @client_notification.user = current_user
+    authorize @client_notification
+    if @client_notification.save
+      redirect_to client_notifications_path,
+        notice: t('crud.created', model: human_model_name)
```
toupeira: I think `make_notice` does the same?

app/controllers/client_notifications_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    @client_notification = ClientNotification.find(params[:id])
+    authorize @client_notification
+  end
+
+  def human_model_name
+    @client_notification.class.model_name.human
+  end
+
+  def client_notification_params
+    params.require(:client_notification).permit(:body, :user_id, :active)
```
toupeira: Looks like `:user_id` is not needed here, the form doesn't contain it and we're always setting it in the controller.

app/models/client_notification.rb
```diff
@@ -0,0 +1,14 @@
+class ClientNotification < ApplicationRecord
+
+  before_save :ensure_exactly_one_active
+
+  default_scope { order(created_at: :desc) }
+  scope :active, -> { find_by(active: true) }
```
toupeira: Scopes are supposed to return chainable `ActiveRecord::Relation`s, in this case I think it would be better to have a class method (`def self.active ...`) (though scopes are actually just class methods too)

app/views/client_notifications/index.html.slim
```diff
@@ -0,0 +1,23 @@
...
+  table.table
+    thead
+      th Erstellt am
+      th Aktiv
+      th Text
+      th Erstellt von
+      th colspan=2
+    tbody
+      - @client_notifications.each do |notification|
+        tr class=(notification.active ? 'bg-success': '') title=(notification.active ? 'Aktiv': '')
```
toupeira: `'bg-success' if notification.active` and `'Aktiv' if notification.active`

app/models/client_notification.rb
```diff
@@ -0,0 +1,14 @@
+class ClientNotification < ApplicationRecord
+
+  before_save :ensure_exactly_one_active
```
toupeira: This should be called in `after_save`, i.e. the other records should only be modified once the active `ClientNotification` has been successfully saved.

Though in practice it doesn't matter much since the saving and before/after callbacks are all run in one DB transaction (except for `after_commit`), so if anything goes wrong all changes will be rolled back.

(so just FYI, no need to really change it I think)

app/policies/client_notification_policy.rb
```diff
@@ -0,0 +1,10 @@
+class ClientNotificationPolicy < ApplicationPolicy
+  # controller action policies
+  alias_method :index?,           :superadmin?
+  alias_method :clients_index?,   :superadmin?
```
toupeira: What's `clients_index` used for? The controller doesn't seem to have an action with that name.

nuritnt: true! I was using it, that only a superadmin can see this button on the clients index, but not necessary indeed

app/controllers/client_notifications_controller.rb
nuritnt: this means, also remove `t.references :user, foreign_key: true` from the migration file and also in the model `belongs_to :user, -> { with_deleted }` ? or only remove it from the required params?

app/models/client_notification.rb
nuritnt: 👍

db/migrate/20171116140737_create_client_notifications.rb
nuritnt: 👍

PR #353, opened 2017-11-16
Features/email-templates

chrysalag:
# [Story in Trello](https://trello.com/c/vFaXKbeV/5-5-email-templates)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Code is reviewed

### Done after the merge?


app/models/email_template.rb
```diff
@@ -0,0 +1,17 @@
+class EmailTemplate < ApplicationRecord
+  before_save :ensure_exactly_one_active_per_kind
+
+  enum kind: { Anmeldung: 0, Probezeit: 1, Begleitung: 2 }
```
casaper: I suggest using enum keys that match the values on the other side. Eg. `Probezeit` -> `probation_period`

The Translation issue, with Enum fields, I've just discovered that it is actually many times easyer as we knew so far. 

If in the form:

```ruby

f.input :kind, collection: EmailTemplate.kinds.keys.map(&:to_sym), ....
```
In the translation yaml:

```yaml
simple_form:
  options:
    email_template:
      kind:
        probation_period: Probezeit 
```
This way it translates perfectly well. It is the good (rather bad) old fuzziness simple form has with the collection keys. It will only translate them if passed as Keys, and it doesn't care if its enum or whatever else. `collection: [:a, :b] ` passed to input will allways look up the options in the simple_form.options.model_name.field.key

casaper: If feasible you could also cherry-pick [this](https://github.com/panter/aoz-003/pull/359/files#diff-1717d3e3835cc67fc19f0620ff5bc59a) here if you want.
I that case you could do exactly like [i did here](https://github.com/panter/aoz-003/blob/f11f43679daa87f37f9fcb8d904cee94ab637f56/app/views/volunteers/_form.html.slim#L7).

app/views/email_templates/new.html.slim
```diff
@@ -0,0 +1,21 @@
+javascript:
+  $( document ).on('turbolinks:render, turbolinks:load', function() {
+    $('#email_template_kind').change(function(){
+      var probezeit = "Liebe/r $[Anrede] $[Name]\n\nWir bitten um Ihre $[Feedback_Link] für $[Begleitung_Titel].\n\nDanke im Voraus!"
```
casaper: I've just discovered, that for our template variables it's probably the best to use `"string to %{replace} in"`.

This is, beacause:

```ruby
"here is %{replace} for me" % { replace: 'text i want to be replaced' }
```

so it will be by far the smoothest and easiest if we did it that way.

app/controllers/volunteer_applications_controller.rb
```diff
@@ -15,8 +15,8 @@ def create
     @volunteer = Volunteer.new(volunteer_params)
     authorize :volunteer_application, :create?
     if @volunteer.save
-      if VolunteerEmail.active_mail.present?
-        VolunteerMailer.welcome_email(@volunteer, VolunteerEmail.active_mail).deliver
+      if EmailTemplate.active_mail.signup.present?
+        VolunteerMailer.welcome_email(@volunteer, EmailTemplate.active_mail.signup.last).deliver
```
toupeira: To avoid duplicate queries I would do instead:

```ruby
email = EmailTemplate.active_mail.signup.last
if email
  ...
```

app/mailers/volunteer_mailer.rb
```diff
@@ -1,8 +1,14 @@
...
     @volunteer = volunteer
-    @volunteer_email = volunteer_email
-    mail(to: volunteer.contact.primary_email, subject: @volunteer_email.subject)
+    @signup_email = signup_email
+    subject = signup_email.subject
+    subject.gsub! '%<anrede>', t("salutation.#{volunteer.salutation}")
+    subject.gsub! '%<name>', "#{volunteer.contact.first_name} #{volunteer.contact.last_name}"
+    body = signup_email.body
+    body.gsub! '%<anrede>', t("salutation.#{volunteer.salutation}")
+    body.gsub! '%<name>', "#{volunteer.contact.first_name} #{volunteer.contact.last_name}"
```
toupeira: This looks very similar to the code in @casaper's [`ReminderMailingVolunteer`](https://github.com/panter/aoz-003/pull/359/files#diff-be29e4532f1777e302a921922fa682e3) model. You can simplify it a bit by defining the hash of replacements (`template_variables = { 'anrede' => ..., ... }`) and then using string formatting (`subject % template_variables`)

Also we should probably use the same placeholder names, i.e. "Anrede" and "Name" (uppercase)

app/views/email_templates/index.html.slim
```diff
@@ -0,0 +1,22 @@
+h1 Alle E-mailvorlagen
```
toupeira: "E-**M**ailvorlagen" (also in other places)

app/views/email_templates/index.html.slim
```diff
@@ -0,0 +1,22 @@
...
+
+= form_navigation_btn :new
+
+.table-responsive
+  table.table
+    thead
+      th Betreff
+      th Textkörper
```
toupeira: "Textkörper" sounds very technical, "Text" is probably enough :)

(also in other places)

app/views/email_templates/new.html.slim
```diff
@@ -0,0 +1,30 @@
+h1= t_title(:new)
+
+p Sie können die folgenden Variablen benützen:
```
toupeira: Maybe "Platzhalter" instead of "Variablen" is clearer

app/views/email_templates/new.html.slim
```diff
@@ -0,0 +1,30 @@
...
+ul
+  li %<anrede>
+  li %<name>
+  li %<feedback-link>
+  li %<assignment-title>
+p Zum Beispeil:
```
toupeira: "Beisp**ie**l""

app/mailers/volunteer_mailer.rb
casaper: I'd definitly agree to use the same placeholder names as the User will see and insert in the frontend.
So In that case it would be german. And why use gsub, if `%{Placeholder}` does the job so much easier?

casaper: The Syntax `%{}` is native, so no gsub is needed.

app/views/email_templates/index.html.slim
```diff
@@ -0,0 +1,22 @@
...
+.table-responsive
+  table.table
+    thead
+      th Betreff
+      th Text
+      th Art
+      th Aktiv
+      th
+    tbody
+      - @email_templates.each do |email_template|
```
nuritnt: some suggestions for the index view & table:
- order of the cells: `Art | Aktiv | Betreff | Text`
- possibility to filter `Art` ?
- you can use `boolean_glyph(email_template.active)` to save some space
- is there a possibility to set the with for the subject or body?
    if you have a big text the subject gets squashed and is userunfriendly to read:
screen shot 2017-11-24 at 12 00 11

app/models/email_template.rb
```diff
@@ -0,0 +1,18 @@
+class EmailTemplate < ApplicationRecord
+  before_save :ensure_exactly_one_active_per_kind
+
+  enum kind: { signup: 0, trial: 1, assignment: 2 }
+  validates :kind, presence: true
+
+  default_scope { order(created_at: :desc) }
+  scope :active_mail, -> { where(active: true) }
```
nuritnt: as @toupeira suggested this also on the `client_notification` I think would make sense to also adapt it here:

app/models/email_template.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def self.kind_collection
+    kinds.keys.map(&:to_sym)
+  end
+
+  def ensure_exactly_one_active_per_kind
+    return unless active && changed.include?('active')
+    EmailTemplate.where(kind: kind, active: true).update(active: false)
```
nuritnt: after you change the active scope to a class method this should also be updated

app/views/email_templates/index.html.slim
chrysalag: There was no request for filtering. 

Regarding the big text possibility, these are just templates, so I doubt there will contain big texts. Big texts are not easily reusable (reusability is the point of having templates).

The icon instead of Ja/Nein, wouldn't save that much of space.

nuritnt: The platform should be consistent. The icons where already used for the `client_notifications`.
Regarding the filter, right it was not requested, but makes it once more consistent and more user friendly - up to you.

app/views/email_templates/new.html.slim
```diff
@@ -0,0 +1,34 @@
...
+
+p Sie können die folgenden Platzhalter benützen:
+ul
+  li %{Anrede}
+  li %{Name}
+  li %{feedback-link}
+  li %{assignment-title}
+p Zum Beispiel:
+p Liebe/r %{Anrede} %{Name}
+p Wir bitten um Ihre %{feedback_link} für %{assignment_title}.
```
nuritnt: is this a typo?
with dash: `%{feedback-link}` & `%{assignment-title}`  line 5 + 6
with underscore: `%{feedback_link}` & `%{assignment_title}`  line 11

and shouldn't `%{assignment_title}` have also a German placeholder name? or will this be changed later anyway?



NIT: `%{Anrede}` & `%{Name}` are capitalised, whereas `%{feedback_link}` & `%{assignment_title}` are not..

app/controllers/volunteer_applications_controller.rb
```diff
@@ -15,9 +15,8 @@ def create
     @volunteer = Volunteer.new(volunteer_params)
     authorize :volunteer_application, :create?
     if @volunteer.save
-      if VolunteerEmail.active_mail.present?
-        VolunteerMailer.welcome_email(@volunteer, VolunteerEmail.active_mail).deliver
-      end
+      email = EmailTemplate.active.signup.last
+      VolunteerMailer.welcome_email(@volunteer, email).deliver if email
```
casaper: What happens if `email` doesn't exist? AFAIS `redirect_to thanks_volunteer_applications_url` will happen anyway, so the volunteer get's no email but a confirmation.

app/models/email_template.rb
```diff
@@ -0,0 +1,21 @@
+class EmailTemplate < ApplicationRecord
+  before_save :ensure_exactly_one_active_per_kind
+
+  enum kind: { signup: 0, trial: 1, assignment: 2 }
+  validates :kind, presence: true
+
+  default_scope { order(active: :desc) }
```
casaper: `default_scope` shouldn't be used

app/views/email_templates/new.html.slim
```diff
@@ -0,0 +1,34 @@
+h1= t_title(:new)
+
+p Sie können die folgenden Platzhalter benützen:
+ul
+  li %{Anrede}
+  li %{Name}
+  li %{feedback-link}
+  li %{assignment-title}
```
casaper: Stay German if possible `%{EinsatzTitel}`

app/views/email_templates/new.html.slim
```diff
@@ -0,0 +1,34 @@
+h1= t_title(:new)
+
+p Sie können die folgenden Platzhalter benützen:
+ul
+  li %{Anrede}
+  li %{Name}
+  li %{feedback-link}
```
casaper: `%{EmailLink}`

casaper: A name that is easy in a hash `{ EmailLink: 'asdfasdf' }` vs `{ 'email-link': 'asdfasdf' }`

casaper: And I would suggest to have something automatic here, as soon as this functionality is zipped with https://github.com/panter/aoz-003/pull/359

app/models/email_template.rb
nuritnt: agree

app/views/email_templates/new.html.slim
chrysalag: Yes, I could temporarily define them in the `EmailTemplate` model.

chrysalag: Yes, I can temporarily define them in the `EmailTemplate` model.

casaper: In the ReminderMailingVolunteer they're right now:
```ruby
[:Anrede, :Einsatz, :Name, :EinsatzStart]
```

chrysalag: Why is the `:EinsatzStart` used for?

app/controllers/volunteer_applications_controller.rb
casaper: I think fixing this would be out of scope. In the edge case, the sending would have failed, the Volunteers application would still have succeeded. So it's an edge case and it wouldn't completely brake things. So I'm taking that back...

app/mailers/volunteer_mailer.rb
```diff
@@ -1,8 +1,16 @@
...
-    mail(to: volunteer.contact.primary_email, subject: @volunteer_email.subject)
+    @signup_email = signup_email
+    template_variables = {
+      Anrede: I18n.t("salutation.#{volunteer.salutation}"),
+      Name: "#{volunteer.contact.first_name} #{volunteer.contact.last_name}"
+    }
+    subject = signup_email.subject
+    @subject = subject % template_variables
+    body = signup_email.body
+    @body = body % template_variables
```
casaper: Nitpicking:
```ruby
@body = signup_email.body % template_variables
```

app/mailers/volunteer_mailer.rb
```diff
@@ -1,8 +1,16 @@
...
     @volunteer = volunteer
-    @volunteer_email = volunteer_email
-    mail(to: volunteer.contact.primary_email, subject: @volunteer_email.subject)
+    @signup_email = signup_email
+    template_variables = {
+      Anrede: I18n.t("salutation.#{volunteer.salutation}"),
+      Name: "#{volunteer.contact.first_name} #{volunteer.contact.last_name}"
+    }
+    subject = signup_email.subject
+    @subject = subject % template_variables
```
casaper: Nitpicking:
```ruby
@subject = signup_email.subject % template_variables
```

PR #359, opened 2017-11-21
EMail Versand Probezeit Feedback

casaper:
# [Story in Trello](https://trello.com/c/Ysg0XPuc/7-3-email-versand-probezeit-feedback)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [x] Code is reviewed

### Done after the merge?


app/models/application_record.rb
```diff
@@ -1,4 +1,15 @@
...
+
+  # translate enum fields value
+  def t_enum(enum_field)
+    I18n.t("activerecord.attributes.#{model_name.i18n_key}."\
+      "#{enum_field.to_s.pluralize}.#{send(enum_field)}")
+  end
+
+  # simple form translation 'simple_form.options.model_name.enum_field.key'
+  def self.enum_collection(enum_field)
+    send(enum_field.to_s.pluralize).keys.map(&:to_sym)
```
toupeira: Always prefer `public_send` if possible

app/views/reminder_mailings/new_half_year.html.slim
```diff
@@ -0,0 +1,3 @@
+h1 Errinnerungs Email Versand erstellen
```
toupeira: Typo "E**r**innerung" (also in other places)

app/models/reminder_mailing_volunteer.rb
```diff
@@ -0,0 +1,32 @@
...
+  belongs_to :reminder_mailing
+  belongs_to :volunteer
+  belongs_to :reminder_mailable, polymorphic: true, optional: true
+
+  private
+
+  def remove_or_process_email_content
+    return destroy if selected && selected.to_i.zero?
+    self.body = reminder_mailing.body % template_variables
+    self.subject = reminder_mailing.subject % template_variables
```
toupeira: Not sure if it makes sense to pre-render these fields and store them in the DB, what happens if one of the template variables change until the mails are sent? Another option would be to render them directly when the actual mails are sent, but it looks like that's not implemented yet.

app/models/reminder_mailing.rb
```diff
@@ -0,0 +1,19 @@
...
+  enum kind: { probation_period: 1, half_year: 0 }
+
+  # setter generates relation to assignment/group_assignment and volunteer in one go
+  def reminder_mailing_volunteers=(reminder_mailable)
+    if [Assignment, GroupAssignment].include? reminder_mailable.first&.class
+      super(reminder_mailable.map do |mailable|
+        ReminderMailingVolunteer.new(volunteer: mailable.volunteer, reminder_mailable: mailable)
+      end)
+    else
+      super(reminder_mailable)
```
toupeira: I'm a bit confused by this, is there a case where the `reminder_mailables` are **not** assignments or group assignments?

app/models/reminder_mailing_volunteer.rb
```diff
@@ -0,0 +1,32 @@
...
+  before_create :remove_or_process_email_content
+
+  belongs_to :reminder_mailing
+  belongs_to :volunteer
+  belongs_to :reminder_mailable, polymorphic: true, optional: true
+
+  private
+
+  def remove_or_process_email_content
+    return destroy if selected && selected.to_i.zero?
```
toupeira: For nested attributes / auto-save associations you can pass the magic field `_destroy` to indicate whether the record should be deleted, i.e. you can remove the `selected` attribute and this line and try just having a checkbox for `_destroy` in the view (not sure if it works as expected in this case)

app/views/reminder_mailings/_form.html.slim
```diff
@@ -0,0 +1,24 @@
...
+    tbody
+      = f.simple_fields_for :reminder_mailing_volunteers do |mailing_volunteer|
+        tr
+          td= mailing_volunteer.input_field :selected, as: :boolean
+          td
+            = mailing_volunteer.input_field :volunteer_id, as: :hidden
+            = mailing_volunteer.input_field :reminder_mailable_id, as: :hidden
+            = mailing_volunteer.input_field :reminder_mailable_type, as: :hidden
+            = mailing_volunteer.object.volunteer.to_s
+          td= mailing_volunteer.object.reminder_mailable.to_s
```
toupeira: `.to_s` is optional in views (`=` already calls it for us)

app/models/reminder_mailing.rb
casaper: ReminderMailing -> ReminderMailingVolunteer -> Assignment or GroupAssignment

But with this setter i can pass:
```ruby
ReminderMailing.new(reminder_mailing_volunteers: [Assignment])
```
so the setter will automaticly build a GroupAssignmentVolunteer record with the relation to the Assignment and Volunteer in one go

casaper: Otherwise it treats it just as a normal ReminderMailingVolunteer record beeing passed

app/models/reminder_mailing_volunteer.rb
casaper: your right. True. I just thought I do it here to fill them as a property. But maybe your right.
But it is not much of a problem tho. I can just convert this into a method, that will spit the substituded subject and body out.

app/views/reminder_mailings/_form.html.slim
casaper: Yea, right.
This is not finite yet anyway tho. I'll need to make sure to have a propper to_label method on both GroupAssignment and Assignment, so that I can display this here.

app/views/reminder_mailings/new_half_year.html.slim
casaper: My bad. thx

test/models/reminder_mailing_volunteer_test.rb
```diff
@@ -0,0 +1,21 @@
...
+class ReminderMailingVolunteerTest < ActiveSupport::TestCase
+  def setup
+    @superadmin = create :user
+    @volunteer = create :volunteer_with_user
+    @assignment_probation = create :assignment, period_start: 7.weeks.ago, volunteer: @volunteer
+  end
+
+  test 'template variables are substituted' do
+    reminder_mailing = ReminderMailing.new(kind: :probation_period, creator: @superadmin,
+      subject: 'hallo %{Anrede} %{Name}', reminder_mailing_volunteers: [@assignment_probation],
```
chrysalag: RuboCop complains if this kind  `%{Anrede}` of form is selected for variables. It suggests this `%<Anrede>` instead.

app/views/reminder_mailings/_form.html.slim
chrysalag: Regarding the selection of the volunteers, there is also a "form" here, that as far as I understand either needs to be adapted based on the new changes or removed. 
https://github.com/panter/aoz-003/blob/develop/app/views/reminders/trial_end.html.slim

test/models/reminder_mailing_volunteer_test.rb
casaper: I have no clue what RuboCop expects me to do here, but as what he suggests to change is rubbish, I'm ignoring him.
This is because:

```ruby
2.4.2 (main):0 > 'hallo %{Anrede} %{Name}' % { Anrede: 'blabla', Name: 'Sonstwas' }
=> "hallo blabla Sonstwas"
2.4.2 (main):0 > 'hallo %<Anrede> %<Name>' % { Anrede: 'blabla', Name: 'Sonstwas' }
ArgumentError: invalid format character - %
2.4.2 (main):0 > format('hallo %<Anrede> %<Name>', Anrede: 'blabla', Name: 'Sonstwas')
ArgumentError: invalid format character - %
2.4.2 (main):0 > format('hallo %<Anrede> %<Name>', {Anrede: 'blabla', Name: 'Sonstwas'})
ArgumentError: invalid format character - %
2.4.2 (main):0 > format('hallo %<Anrede> %<Name>', {Anrede: 'blabla', Name: 'Sonstwas'})
ArgumentError: invalid format character - %
```

What rubocop suggests looks uglier, works uglier, .... well it just is 💩 

And the solution is simple:

```yaml
Style/FormatStringToken:
  EnforcedStyle: template
```

http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/FormatStringToken

casaper: And the rubocop documentation doesn't say that this would be any worse than the other solution. Sometimes Rubocop could really give the reason why x should be enforced y ...

app/models/department.rb
```diff
@@ -11,6 +11,10 @@ class Department < ApplicationRecord
 
   validates :contact, presence: true
 
+  def to_label
```
nuritnt: why do we have a method with different names, but does the same?

app/controllers/reminder_mailings_controller.rb
```diff
@@ -0,0 +1,87 @@
+class ReminderMailingsController < ApplicationController
+  before_action :set_reminder_mailing, only: [:show, :edit, :update, :destroy, :initiate_mailing]
+
+  def index
+    authorize ReminderMailing
+    @reminder_mailings = ReminderMailing.order('created_at desc').all
```
nuritnt: why do we need the ordering here instead of how we used it like before with a scope?

app/controllers/reminder_mailings_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def edit; end
+
+  def initiate_mailing
+    if @reminder_mailing.sending_triggered
+      return redirect_to reminder_mailings_path, notice: 'Dieses Mailing wurde bereits versandt'
+    end
+    @reminder_mailing.reminder_mailing_volunteers.each do |mailing_volunteer|
+      VolunteerMailer.probation_period_reminder(mailing_volunteer).deliver_later
```
nuritnt: what does `deliver_later` do? and where does it come from?

app/models/concerns/group_assignment_and_assignment_common.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    scope :need_probation_period_reminder_mailing, lambda {
+      started_ca_six_weeks_ago.no_reminder_mailing
+    }
+
+    scope :with_reminder_mailing_kind, lambda { |kind_number|
+      loj_mailings.where('reminder_mailings.kind = ?', kind_number)
+    }
+
+    scope :with_reminder_mailing_kind, lambda { |kind_number|
```
nuritnt: this scope is a duplicate from line 54

app/models/concerns/group_assignment_and_assignment_common.rb
```diff
@@ -0,0 +1,87 @@
...
+      loj_mailings
+        .where('reminder_mailings.kind != 1 OR reminder_mailing_volunteers.id IS NULL')
+    }
+
+    def started_six_months_ago?
+      period_start < 6.months.ago
+    end
+
+    def started_ca_six_weeks_ago?
+      period_start < 6.weeks.ago && period_start > 8.weeks.ago
```
nuritnt: why do we need again this and the method above if we have these already defined as scopes? ^^

app/views/reminder_mailings/_reminder_mailing.html.slim
```diff
@@ -0,0 +1,12 @@
...
+  td= link_to reminder_mailing.creator.to_label, reminder_mailing.creator
+  td
+    - if reminder_mailing.sending_triggered
+      'Übermittelt
+    - else
+      'Nicht versandt
+  td= l(reminder_mailing.created_at.to_date)
+  td
+    = link_to 'Anzeigen', reminder_mailing, class: 'index-action-link'
+    = link_to 'Bearbeiten', edit_reminder_mailing_path(reminder_mailing), class: 'index-action-link'
```
nuritnt: `- if !reminder_mailing.sending_triggered` 
should one be able to edit a mail which is anyway already sent out?

app/views/volunteer_mailer/probation_period_reminder.html.slim
```diff
@@ -0,0 +1,3 @@
+h1= @subject
+div= markdown(@body)
+div Some bogus footer
```
nuritnt: whats this? 😆

db/seeds.rb
```diff
@@ -33,125 +20,146 @@ def make_relatives
...
-    Journal.new(
-      body: Faker::Lorem.sentence(rand(2..5)),
-      user: user,
-      category: random_category
+      category: Journal::CATEGORIES.sample
     )
   ]
   user.clients << Array.new(2).map do
-    FactoryBot.create :client_seed, journals: journals, relatives: make_relatives
+    FactoryBot.create :client_seed, journals: journals, relatives: make_relatives, user: user
```
nuritnt: looks like journals seed didn't work...

app/controllers/reminder_mailings_controller.rb
nuritnt: clarified

test/mailers/volunteer_mailer_test.rb
```diff
@@ -25,4 +30,18 @@ def setup
...
   end
+
+  test 'probation_period_mailer' do
+    _, _, group_assignments = create_group_offer_entity(
+      nil, 7.weeks.ago, nil, create(:volunteer_with_user), create(:volunteer_with_user)
+    )
+    assignment = make_assignment(start_date: 7.weeks.ago)
+    mailing = create_probation_mailing(*group_assignments, assignment)
+    mailing.reminder_mailing_volunteers.each do |rmv|
+      mailer = VolunteerMailer.probation_period_reminder(rmv).deliver
```
nuritnt: I know you are using in the model `deliver_later` to not make the user wait, but here you are using only `deliver`. wouldn't it make sense to use the same method for testing?

app/controllers/reminder_mailings_controller.rb
```diff
@@ -45,7 +45,12 @@ def create
     end
   end
 
-  def edit; end
+  def edit
+    if @reminder_mailing.sending_triggered
+      redirect_back(fallback_location: reminder_mailing_path(@reminder_mailing),
+        notice: 'Wenn das mailing bereits versendet wurde, kann es nicht mehr geändert werden.')
```
nuritnt: NIT: Mailing

test/mailers/volunteer_mailer_test.rb
casaper: nope, because if I use deliver_later in the test I cannot know if and when the email will be delivered. And as I want to know what the mail is like I want to get it delivered now. This isn't about testing deliver_later or deliver. Its for my own funcionality

nuritnt: aaaah 💡  yes makes sense

app/views/reminder_mailings/_form.html.slim
```diff
@@ -0,0 +1,38 @@
...
+              td.click-selecting= mailing_volunteer.object.reminder_mailable.to_label
+              td
+                - if mailing_volunteer.object.group_assignment?
+                  = link_to 'Einsatz Info', mailing_volunteer.object.reminder_mailable.group_offer, target: '_blank'
+                - else
+                  = link_to 'Einsatz Info', mailing_volunteer.object.reminder_mailable, target: '_blank'
+
+    .col-xs-12.col-md-6
+      h3 Email Vorlage bearbeiten
+      = f.input :subject
```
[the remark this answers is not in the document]
casaper: That is done from the controller.

db/migrate/20171121124734_create_reminder_mailings.rb
```diff
@@ -0,0 +1,15 @@
+class CreateReminderMailings < ActiveRecord::Migration[5.1]
+  def change
+    create_table :reminder_mailings do |t|
+      t.references :creator, references: :users, index: true
+      t.text :body
+      t.string :subject
+      t.integer :kind, default: 0
+      t.boolean :sending_triggered, default: false
```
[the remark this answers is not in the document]
casaper: as updateing the model is not possible after mailing is triggered, the sending_triggered with updated_at deliveres this information allready.

app/views/volunteer_mailer/probation_period_reminder.html.slim
```diff
@@ -0,0 +1,2 @@
+h1= @subject
```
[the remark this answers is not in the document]
casaper: no, that doesn't make sense at all, sorry. Because I'd need to partialize the view for web and it would make the thing totally unreadable. because the mailing only wants two elements `h1` and `div`.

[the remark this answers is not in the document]
casaper: ah, now I see. The file `app/views/reminder_mailings/probation_period_reminder.html.slim` is useless where it is. I'll move it. Thanks!

db/migrate/20171121124734_create_reminder_mailings.rb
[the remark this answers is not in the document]
casaper: But maybe your right, I could display that information.

app/controllers/reminder_mailings_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      redirect_to @reminder_mailing, make_notice
+    else
+      render "new_#{@reminder_mailing.kind}".to_sym
+    end
+  end
+
+  def edit
+    if @reminder_mailing.sending_triggered
+      redirect_back(fallback_location: reminder_mailing_path(@reminder_mailing),
+        notice: 'Wenn das mailing bereits versendet wurde, kann es nicht mehr geändert werden.')
```
toupeira: "**M**ailing"? Or maybe a more German word? ;-)

app/controllers/reminder_mailings_controller.rb
toupeira: It's available on all mailers and enqueues a job to send the mail rather than sending it directly, see http://api.rubyonrails.org/classes/ActionMailer/MessageDelivery.html#method-i-deliver_later

Note that the default job queue in Rails will lose jobs after a server restart, for mails like this it should be ok but for more important jobs something like Sidekiq should be used.

app/models/concerns/group_assignment_and_assignment_common.rb
toupeira: The scopes are for filtering all records on the SQL level, while the methods are for checking individual objects on the Ruby level. Does that make sense?

app/models/department.rb
```diff
@@ -11,6 +11,10 @@ class Department < ApplicationRecord
...
+    contact.last_name
+  end
+
```
toupeira: Why not reuse the existing `.to_s`?

app/models/reminder_mailing.rb
```diff
@@ -0,0 +1,58 @@
...
+    errors.add(:volunteers, 'Es muss mindestens ein Freiwilliger ausgewählt sein.')
+  end
+
+  def mailing_not_sent
+    if sending_triggered && !will_save_change_to_sending_triggered?
+      errors.add(:already_sent, 'Wenn das mailing bereits versendet wurde, kann es nicht mehr '\
+        'geändert werden.')
+      false
+    else
+      true
```
toupeira: Note that it doesn't matter if this method returns `true` or `false`, it only needs to add errors.

app/views/application/_navigation.html.slim
```diff
@@ -28,10 +28,8 @@ nav.navbar.navbar-top.hidden-print
             li= link_to 'Gruppenangebote', group_offers_path
           - if policy(PerformanceReport).index?
             li= link_to 'Kennzahlen', performance_reports_path
-          - if policy(Reminder).index?
-            li= link_to 'Erinnerungen', reminders_path
-          - if policy(Reminder).trial_end?
-            li = link_to 'Probezeit Erinnerungen', trial_end_reminders_path
+          - if policy(ReminderMailing).index?
+            li = link_to 'Errinnerungs Mailings', reminder_mailings_path
```
toupeira: "E**r**innerungs-Mailings"

app/views/reminder_mailings/_form.html.slim
```diff
@@ -0,0 +1,37 @@
+= simple_form_for @reminder_mailing do |f|
+  = simple_error_notice f
+  = f.input_field :kind, value: @reminder_mailing.kind, as: :hidden
+
+  = render 'form_submit_back', f: f
+
+  .row
+    .col-xs-12.col-md-6
+      h3 Frewillige auswählen
+      - if f.error_notification.present? && f.object.errors.messages[:volunteers].present?
```
toupeira: `f.object.errors.include?(:volunteers)`

app/views/reminder_mailings/_section_navigation.html.slim
```diff
@@ -0,0 +1,7 @@
+nav.navbar.section-navigation.hidden-print
+  ul.section-nav-ul.list-unstyled
+    li= button_link 'Email Errinnerungs Übersicht', reminder_mailings_path, dimension: :sm
+    li= button_link 'Probezeit Errinnerung erstellen', new_probation_period_reminder_mailings_path, dimension: :sm
+    li= button_link 'Halbjahres Errinnerung erstellen', new_half_year_reminder_mailings_path, dimension: :sm
```
toupeira: "E**r**innerung" (other places too)

app/views/reminder_mailings/show.html.slim
```diff
@@ -0,0 +1,54 @@
...
+            = link_to mailing_volunteer.reminder_mailable.group_offer.to_label, mailing_volunteer.reminder_mailable.group_offer
+
+
+nav.navbar.hidden-print
+  ul.list-inline
+    - if !@reminder_mailing.sending_triggered
+      li= button_link 'Emails versenden', initiate_mailing_reminder_mailing_path(@reminder_mailing)
+      li= button_link 'Ändern', edit_reminder_mailing_path(@reminder_mailing)
+      li= link_to 'Löschen', reminder_mailing_path(@reminder_mailing), confirm_deleting(@reminder_mailing, 'btn btn-default')
+      li= button_link 'Zurück', edit_reminder_mailing_path(@reminder_mailing)
```
toupeira: Do we need both the "Ändern" and "Zurück" buttons? They both point to the same page.

test/application_system_test_case.rb
```diff
@@ -63,4 +63,16 @@ def fill_autocomplete(name, options = {})
...
+  def any_checked?(selector)
+    page.find_all(selector).reduce do |a, b|
+      a.checked? || b.checked?
+    end
+  end
+
+  def all_checked?(selector)
+    page.find_all(selector).reduce do |a, b|
+      a.checked? && b.checked?
+    end
```
toupeira: Not sure why you're using `reduce` here, you should be able to just do `.any?(&:checked)` and `.all?(&:checked)`

test/mailers/volunteer_mailer_test.rb
```diff
@@ -1,6 +1,11 @@
 require 'test_helper'
+require 'utility/reminder_mailing_builder'
+require 'utility/group_offer_and_assignment'
```
toupeira: Why not load these in `test_helper`?

test/mailers/volunteer_mailer_test.rb
```diff
@@ -1,6 +1,11 @@
...
 
 class VolunteerMailerTest < ActionMailer::TestCase
+  include ReminderMailingBuilder
+  include GroupOfferAndAssignment
```
toupeira: You could also move this into `test_helper`, something like:

```ruby
class ActionMailer::TestCase
  include ...
end
```

test/models/reminder_mailing_volunteer_test.rb
toupeira: It's based on the [Ruby style guide](https://github.com/bbatsov/ruby-style-guide#named-format-tokens):

> When using named format string tokens, favor `%<name>s` over `%{name}` because it encodes information about the type of the value.
>
>  ```ruby
>  # bad
>  format('Hello, %{name}', name: 'John')
>
>  # good
>  format('Hello, %<name>s', name: 'John')
>  ```

With `%<>` you also need to specify the type as a suffix (e.g. `%<Anrede>s`), which is why you got that error.

I'm happy with either style, as long as it's consistent ;-)

test/system/reminder_mailings_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    assert page.has_link? 'Einsatz Info', href: group_offer_path(@group_assignment.group_offer)
+
+    # All checkboxes are not checked?
+    assert_equal false, page.find_all(
+      'input[name^="reminder_mailing[reminder_mailing_volunteers_attributes]"]'
+    ).reduce { |a, b| a.checked? || b.checked? }
+
+    find('td', text: @assignment.to_label).click
+    # at least one checkbox is checked?
+    assert_equal true, any_checked?(
```
toupeira: `assert any_checked?(...`

test/system/reminder_mailings_test.rb
```diff
@@ -0,0 +1,82 @@
...
+      'input[name^="reminder_mailing[reminder_mailing_volunteers_attributes]"]')
+    # not all checkboxes are checked
+    assert_equal false, all_checked?(
```
toupeira: `refute all_checked?(...`

app/models/email_template.rb
```diff
@@ -1,7 +1,7 @@
 class EmailTemplate < ApplicationRecord
   before_save :ensure_exactly_one_active_per_kind
 
-  enum kind: { signup: 0, trial: 1, assignment: 2 }
+  enum kind: { signup: 0, trial: 1, assignment: 2, half_year: 3 }
```
chrysalag: I am sorry but I am a bit lost, what is the difference between the `assignment:` and the `half year:`?
Maybe 
`enum kind: { signup: 0, trial: 1, assignment: 2, group_assignment: 3 }` or something like that could make more sense, since every half year both assignment and group assignment have feedback and hours.

app/views/reminder_mailings/_section_navigation.html.slim
```diff
@@ -0,0 +1,7 @@
...
+
```
chrysalag: empty lines can be removed

app/assets/stylesheets/_bootstrap_overwrites.scss
```diff
@@ -1,16 +1,47 @@
...
+$navbar-height: 40px;
 
+// Non bootstrap lenghts
 $header-height: 70px;
-$navbar-height: 40px;
 $sidebar-width: 220px;
 $footer-height: 30px;
 
+$a4-short-side: 210mm;
+$a4-long-side: 297mm;
```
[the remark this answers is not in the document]
casaper: well, height and width are dependant on wich way you turn the page... so no.... 😉

app/assets/javascripts/mailings_select_all.es6
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  const selectedBoxes = $('input.boolean[id$="_selected"]');
+  selectedBoxes.on('change', ({target}) => {
+    const row = $(target).closest('tr');
+    if (target.checked) {
+      row.addClass('mailing-selected');
+    } else
+      row.removeClass('mailing-selected');
+    }
```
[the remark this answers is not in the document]
casaper: what if for some reason the element is allready `tr.mailing-selected` and the checked is false?
In that case toggle would remove it, despite it should be there.

app/assets/javascripts/mailings_select_all.es6
```diff
@@ -0,0 +1,37 @@
+$(() => $(document).on('turbolinks:render, turbolinks:load', () => {
+
+  $('.reminder-select .click-selecting').on('click', event => {
+    event.stopPropagation();
+    const row = $(event.target).closest('tr');
+    const selectBox = $(`#reminder_mailing_reminder_mailing_volunteers_attributes_${row.data().index}_selected`)[0];
+    if (selectBox.checked) {
+      row.removeClass('mailing-selected');
+      $(selectBox).prop('checked', false);
```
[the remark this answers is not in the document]
casaper: true. 👍

test/system/reminder_mailings_test.rb
casaper:
```ruby
class ApplicationSystemTestCase
# ...


  def any_checked?(selector)
    page.find_all(selector).reduce do |a, b|
      a.checked? || b.checked?
    end
  end

  def all_checked?(selector)
    page.find_all(selector).reduce do |a, b|
      a.checked? && b.checked?
    end
  end
end
```

test/system/reminder_mailings_test.rb
casaper: see above.

app/models/email_template.rb
casaper: Ah, afair the half year will not be just about assignments. I didn't know what assignment would be, so thats why I added this.

chrysalag: The assignment had the same reasoning as the one you have for the half year.

casaper:
@toupeira 

> It's based on the Ruby style guide:

>>   When using named format string tokens, favor %<name>s over %{name} because it encodes 
>>   information about the type of the value.

>>       # bad
>>       format('Hello, %{name}', name: 'John')

>>       # good
>>      format('Hello, %<name>s', name: 'John')

>  With %<> you also need to specify the type as a suffix (e.g. %<Anrede>s), which is why you got that error.

Ah, That's why. Now Rubocop makes quite more sense. But they're style guide explanation is still Rubosh in my view because they could have stated that reason there.... 🐢 

> I'm happy with either style, as long as it's consistent ;-)

Anyhow, I'll stick with `%{VarName}` because we're going to expect our users with those, and we can do it with `%{}` but I would definitely not expect them to do a `%<VarName>s` as this is not feasible to a user.

test/application_system_test_case.rb
casaper: As soon as you made me aware of this, I'm sure I want to use `.all?(&:checked?)` and `.any?(&:checked?)`! 🧀 🐭

app/views/reminder_mailings/show.html.slim
casaper: Zurück will be dropped.

app/views/reminder_mailings/_form.html.slim
```diff
@@ -0,0 +1,62 @@
...
+            = "Wir bitten um Ihre Rückmeldung für:\r\n%{Einsatz}.\r\n"
+            = "Danke im Voraus!\r\n"
+          p Resultat:
+          pre
+            = "Guten Tag Frau Marie Muster\r\n"
+            = "Wir bitten um Ihre Rückmeldung für:\r\nJogakurs - Zentrum Juch\r\n"
+            = "Danke im Voraus!\r\n"
+    .col-xs-12.col-md-8
+      fieldset
+        legend Betreff
```
nuritnt: since `Subject`& `Body` are required, and you have the legend plus the labels(which are shown yet still in english), I would suggest to go for this:

```
fieldset
    legend * Betreff
    = f.input :subject, label: false
fieldset
    legend * Text
    = f.input :body, label: false
```

then you don't have twice the same labels, and still the User knows, that they're required

app/views/reminder_mailings/show.html.slim
```diff
@@ -0,0 +1,53 @@
...
+dl.dl-horizontal
+  dt Art
+  dd= @reminder_mailing.t_enum(:kind)
+  dt Erstellt am
+  dd= l(@reminder_mailing.created_at.to_date)
+  dt Status
+  - if @reminder_mailing.sending_triggered
+    dd Übermittelt
+  - else
+    dd Nicht versandt
```
nuritnt: for enhancing readability this view, since a lot of information are shown, maybe a `hr` would be helpful here

app/views/reminder_mailings/show.html.slim
```diff
@@ -0,0 +1,53 @@
...
+h3 Email Beispiel
+- example = @reminder_mailing.reminder_mailing_volunteers.sample.process_template
+h4 Betreff
+.mailing-body-preview= example[:subject]
+h4 Text
+.mailing-body-preview= markdown(example[:body])
+
```
nuritnt: for enhancing readability this view, since a lot of information are shown, maybe a `hr` would be helpful here

app/views/reminder_mailings/show.html.slim
```diff
@@ -0,0 +1,53 @@
+= render 'section_navigation'
+
+h1 #{@reminder_mailing.t_enum(:kind)} Erinnerungs-Email
```
nuritnt: please check the rendering of this file, since the footer is not sticky at the bottom anymore on this view

app/views/reminder_mailings/_form.html.slim
```diff
@@ -0,0 +1,62 @@
...
+        .panel-heading
+          strong Erklärung
+        .panel-body
+          'Sie können die folgenden Platzhalter benützen:
+
+        ul#template-variables.list-group
+          - ReminderMailing::TEMPLATE_VARNAMES.each do |var_name|
+            li.list-group-item %{#{var_name}}
+
+        .panel-body
```
chrysalag: It seems the instructions are the same with the email template form, so maybe there could be accessible through a template view.

app/views/reminder_mailings/show.html.slim
```diff
@@ -0,0 +1,53 @@
...
+h2 Versand Einzelheiten
+
+dl.dl-horizontal
+  dt Art
+  dd= @reminder_mailing.t_enum(:kind)
+  dt Erstellt am
+  dd= l(@reminder_mailing.created_at.to_date)
+  dt Status
+  - if @reminder_mailing.sending_triggered
+    dd Übermittelt
```
chrysalag: sent date could also be displayed here

app/views/reminder_mailings/_reminder_mailing.html.slim
```diff
@@ -0,0 +1,13 @@
+tr
+  td= reminder_mailing.t_enum(:kind)
+  td= link_to reminder_mailing.creator.to_label, reminder_mailing.creator
+  td
+    - if reminder_mailing.sending_triggered
+      'Übermittelt am #{l(reminder_mailing.updated_at.to_date)}
+    - else
+      'Nicht versandt
```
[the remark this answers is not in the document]
casaper: this is a form of syntax in slim language:
```slim
p
   'Abcede 
   ' defghij
p
   = 'Abcede'
   = ' defghij'
```

They both result in the same:
```html
<p>Abcede defghij</p>
<p>Abcede defghij</p>
```

nuritnt:
@casaper ah and one thing: If I enter a variable in the template which we don't have defined, the application crashes - which of course makes sense, but it would be helpful for the Users to have a error message that this `variable is not supported`, instead of having the application crashing.

app/views/email_templates/_form.html.slim
```diff
@@ -1,11 +1,53 @@
...
-
-  .row
-    .col-xs-12= f.button :submit, class: 'btn btn-success'
+  .row.m-t-30
+    .col-xs-12.col-md-4
+      .panel.panel-default
+        .panel-heading
+          strong Erklärung
+        .panel-body
+          'Sie können die folgenden Platzhalter benützen:
```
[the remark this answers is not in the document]
casaper: This is due to makrdown rendering. Markdown wants you to make a blank line to get a new paragraph. 
Markdown input:
```markdown
%{Anrede} %{Name}

You better not touch my gal, or I'll pop you in the kisser, pal

%{Einsatz} %{EinsatzStart} %{FeedbackLink}


You better not touch my gal, or I'll pop you in the kisser, pal
%{Einsatz} %{EinsatzStart} %{FeedbackLink}
```

HTML result:
```html
<div class="mailing-body-preview"><p>Herr Salome Jander</p>

<p>You better not touch my gal, or I'll pop you in the kisser, pal</p>

<p>Begleitung - Marisa, Enrico - 08.10.2017 08.10.2017 </p>

<p>You better not touch my gal, or I'll pop you in the kisser, pal
Begleitung - Marisa, Enrico - 08.10.2017 08.10.2017 </p>
</div>
```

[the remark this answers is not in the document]
nuritnt: then just add a comment to anleitungen, because the users probably wont know this and then they don't need to worry about line breaks in the preview.

app/views/reminder_mailings/_form.html.slim
casaper: They're not exactly the same.

app/models/reminder_mailing_volunteer.rb
```diff
@@ -0,0 +1,64 @@
...
+    delete unless @selected
+  end
+
+  def template_variables
+    template_hash = ReminderMailing::TEMPLATE_VARNAMES.map { |varname| [varname, ''] }.to_h
+    template_hash[:Anrede] = I18n.t("salutation.#{volunteer.salutation}")
+    template_hash[:Einsatz] = mailable_to_label
+    template_hash[:Name] = volunteer.contact.natural_name
+    template_hash[:EinsatzStart] = I18n.l(reminder_mailable.period_start)
+    # TODO: this path will need to come with the TrialFeedback implementation
```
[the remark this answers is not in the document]
casaper: So you want to wait for this implementation with link and everything?

app/assets/stylesheets/_bootstrap_custom.scss
```diff
@@ -0,0 +1,59 @@
...
+@import 'bootstrap/dropdowns';
+@import 'bootstrap/button-groups';
+@import 'bootstrap/input-groups';
+@import 'bootstrap/navs';
+@import 'bootstrap/navbar';
+@import 'bootstrap/breadcrumbs';
+@import 'bootstrap/pagination';
+@import 'bootstrap/pager';
+@import 'bootstrap/labels';
+// @import 'bootstrap/badges';
```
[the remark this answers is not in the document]
casaper: Nope, this one has to stay!
This is for having documented wich bootstrap components are excluded.

app/models/reminder_mailing_volunteer.rb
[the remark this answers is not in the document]
chrysalag: We could also add a story so as not to forget about it.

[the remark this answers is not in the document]
casaper: @chrysalag yes, that could be a way, because this involves more than just a link. It needs logic within the TrialFeedback and the ReminderMailing + ReminderMailingVolunteer models.

app/views/email_templates/new.html.slim
```diff
@@ -1,32 +1,29 @@
...
-
 javascript:
-  $(document).on('turbolinks:render, turbolinks:load', function() {
-    $('#email_template_kind').change(function(){
-      $("#email_template_subject").text("");
-      $("#email_template_body").text("");
-      var kind = $("#email_template_kind").val();
-      var subject = $("input[TYPE=HIDDEN]#subjects_" +kind).val();
-      var body = $("input[TYPE=HIDDEN]#bodies_" +kind).val();
+  $(document).on('turbolinks:render, turbolinks:load', function () {
```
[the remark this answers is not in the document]
casaper: So we will not provide initial Templates for each kind? I would rather consider to provide initial Templates for each kind, so that cannot happen, plus it would be just way more userfriendly (user sees how it could/should be done)

casaper:
@nuritnt 
> ah and one thing: If I enter a variable in the template which we don't have defined, the application crashes - which of course makes sense, but it would be helpful for the Users to have a error message that this variable is not supported, instead of having the application crashing.

I made now a somewhat temporary solution that simply removes these from the text in case there is any wrong ones. Need some expertise with @toupeira or @alesrosina or some other senior. If I see it, it could also be a breach.... I'm not sure.

PR #361, opened 2017-11-22
Features/probezeit-feedbacks

chrysalag:
# [Story-1 in Trello](https://trello.com/c/jJFeJ1SH/8-3-probezeit-feedback-erfassen)
# [Story-2 in Trello](https://trello.com/c/3AFvKXCX/9-3-%C3%BCbersicht-probezeit-feedbacks)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [x] Seeds created
* [ ] Code is reviewed

### Done after the merge?


app/views/trial_feedbacks/show.html.slim
```diff
@@ -0,0 +1,39 @@
...
+          td Erstellt von
+          td= link_to @trial_feedback.author.full_name, @trial_feedback.author
+        tr
+          td Eingesehen von
+          td
+            - if @trial_feedback.reviewer
+              = link_to @trial_feedback.reviewer.full_name, @trial_feedback.reviewer if @trial_feedback.reviewer
+            - else
+              = button_to 'Angeschaut', polymorphic_path([@volunteer, @trial_feedback.feedbackable, @trial_feedback],
+                { trial_feedback: { volunteer_id: @trial_feedback.volunteer.id, feedbackable_id: @trial_feedback.feedbackable.id, id: @trial_feedback.id } }),
```
nuritnt: could we put this maybe in a helper method? 3 lines for a button action makes it hard to read

app/views/trial_feedbacks/show.html.slim
```diff
@@ -0,0 +1,39 @@
...
+      - if policy(TrialFeedback).superadmin_priviledges?
+        tr
+          td= t_model(Volunteer)
+          td= link_to @trial_feedback.volunteer.contact.full_name, volunteer_path(@trial_feedback.volunteer)
+      - if @trial_feedback.assignment?
+        tr
+          td= t_model(Assignment)
+          td= link_to @trial_feedback.feedbackable.to_label, assignment_path(@trial_feedback.feedbackable)
+        tr
+          td= t_model(Client)
```
casaper: Is a Link to Client and Volunteer really needed, because a link to the Assignment could also do the job? Or not?
If not, it could be much slicker with:
```slim
tr
  td= t_model(@trial_feedback.feedbackable.class)
  td= link_to @trial_feedback.feedbackable.to_label, url_for(@trial_feedback.feedbackable)
```

db/migrate/20171122175340_create_trial_feedbacks.rb
```diff
@@ -0,0 +1,15 @@
+class CreateTrialFeedbacks < ActiveRecord::Migration[5.1]
+  def change
+    create_table :trial_feedbacks do |t|
+      t.text :body
+      t.integer :feedbackable_id
+      t.string :feedbackable_type
+      t.index [:feedbackable_id, :feedbackable_type]
```
casaper:
```ruby
t.references :feedbackable, polymorphic: true, index: true
```

test/factories/trial_feedbacks.rb
```diff
@@ -0,0 +1,8 @@
+FactoryBot.define do
+  factory :trial_feedback do
+    association :feedbackable, factory: :assignment
+    volunteer
```
casaper: this one requires an after build method for the factory because otherwise, you're going to get three different volunteers, one for the assignment, the user and the trial_feedback each. But it should be the same.

```ruby
after(:build) do |trial_feedback|
  trial_feedback.volunteer: trial_feedback.assignment.volunteer
  trial_feedback.author: create(:user_volunteer, volunteer: trial_feedback.assignment.volunteer)
end
```

nuritnt: So we checked so far, and this is not the case. The factory is called by the seeds and the tests, the volunteer, feedbackable and author set by the factory the are overwritten by them. (cla&tnt)

app/views/trial_feedbacks/show.html.slim
nuritnt: I think that's the consistent behaviour they are looking for to access the related persons profile. That's why I would leave it as it is.

app/models/trial_feedback.rb
```diff
@@ -0,0 +1,28 @@
+class TrialFeedback < ApplicationRecord
+  attr_reader :trial_feedbackable_id_and_type
```
toupeira: This is unnecessary since you're defining your own reader method below.

app/models/trial_feedback.rb
```diff
@@ -0,0 +1,28 @@
...
+    trial_feedbackable_type == 'GroupOffer'
+  end
+
+  def trial_feedbackable_id_and_type=(id_and_type)
+    self.trial_feedbackable_id, self.trial_feedbackable_type = id_and_type.split(',', 2)
+  end
+
+  def trial_feedbackable_id_and_type
+    "#{trial_feedbackable_id},#{trial_feedbackable_type}"
+  end
```
toupeira: This is the third time we're using this `_id_and_type` pattern, would be a good time to extract it into a concern :)

app/models/user.rb
```diff
@@ -14,6 +14,8 @@ class User < ApplicationRecord
   has_many :journals
   has_many :assignments, inverse_of: 'creator'
   has_many :feedbacks, inverse_of: 'author'
+  has_many :trial_feedbacks, inverse_of: 'author'
+  has_many :trial_feedbacks, inverse_of: 'reviewer'
```
toupeira: You can't define the same association twice, the second definition will replace the first one.

app/models/trial_feedback.rb
nuritnt: Yes you are right. 
We agreed on prefering to move stuff to concerns after the big open pull requests are once merged to avoid more complications.

PR #364, opened 2017-11-29
Features/feedback-and-hour-display

chrysalag:
# [Story in Trello](https://trello.com/c/skezYRxs/4-5-stundenerfassung-feedback-muss-best%C3%A4tigt-werden-k%C3%B6nnen)

### Done for approval?

* [x] Story under test
* [x] Mobile works
* [ ] Seeds created
* [ ] Code is reviewed

### Done after the merge?


app/models/assignment.rb
```diff
@@ -70,4 +70,20 @@ def to_label
...
+  def hours_since_last_submitted
+    if submitted_at.present?
+      hours.where('created_at > ?', submitted_at)
+    else
+      hours
+    end
+  end
+
+  def feedbacks_since_last_submitted
+    if submitted_at.present? && volunteer.user
```
casaper: Why do you check for the volunteer having a user here? If the volunteer has no user the query on `where(author: volunteer.user)` will simply be empty, because `volunteer.user` is `nil` in that case.

app/models/assignment.rb
```diff
@@ -70,4 +70,20 @@ def to_label
...
+      feedbacks.where('created_at > ? AND author_id = ?', submitted_at, volunteer.user.id)
```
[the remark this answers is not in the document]
casaper: And the Feedbacks scope can check for `submitted_at` in its relation for itself, because it can reach the needed information:

```ruby
# Feedback
scope :some_name, lambda {
  if feedbackable.submitted_at.present?
    where('created_at > ?', feedbackable.submitted_at).where(author: volunteer.user)
  else
    where(author: volunteer.user)
  end
}

# AssignmentController
@assignment.feedbacks.some_name
```

app/models/group_assignment.rb
```diff
@@ -43,6 +43,22 @@ def to_label
     label
   end
 
+  def hours_since_last_submitted
+    if submitted_at.present?
+      group_offer.hours.where('created_at > ? AND volunteer_id = ?', submitted_at, volunteer)
```
casaper: Same as above. Put that into `Hour` scopes, and split up the parts in chained wheres. The `Hour` model has the information needed ready as well...

app/models/group_assignment.rb
```diff
@@ -43,6 +43,22 @@ def to_label
...
+    else
+      group_offer.hours.where(volunteer: volunteer)
+    end
+  end
+
+  def feedbacks_since_last_submitted
+    if submitted_at.present? && volunteer.user
+      group_offer.feedbacks.where('created_at > ? AND author_id = ?', submitted_at, volunteer.user.id)
```
casaper: Same as in `Assignment`

app/controllers/feedbacks_controller.rb
```diff
@@ -61,6 +62,19 @@ def set_feedback
     authorize @feedback
   end
 
+  def create_redirect
+    if session[:request_url].include?('last_submitted_hours_and_feedbacks')
```
casaper: I'd suggest to simplify this by having it just to return the path, and not the `redirect_to`. So:
```ruby
def create
  # ...
  redirect_to create_redirect, make_notice
```

app/controllers/feedbacks_controller.rb
```diff
@@ -61,6 +62,19 @@ def set_feedback
...
+      if @feedbackable.class == Assignment
```
casaper: `@feedback` has the `group_offer?` and `assignment?` methods, that will do what you want a bit slimmer.

app/controllers/feedbacks_controller.rb
```diff
@@ -61,6 +62,19 @@ def set_feedback
...
+        redirect_to last_submitted_hours_and_feedbacks_assignment_path(@feedbackable), make_notice
+      else
+        group_assignment = @feedbackable.group_assignments.where(volunteer: @volunteer).last
+        redirect_to last_submitted_hours_and_feedbacks_group_assignment_path(group_assignment), make_notice
```
casaper: It would be easier to do:
```ruby
@feedbackable = @feedbackable.group_assignments.where(volunteer: @volunteer).last if @feedback.group_offer?
polymorphic_path(@feedbackable, action: :last_submitted_hours_and_feedbacks)

File: app/controllers/assignments_controller.rb [hunk 8]
```diff
@@ -1,5 +1,6 @@
 class AssignmentsController < ApplicationController
-  before_action :set_assignment, only: [:show, :edit, :update, :destroy]
+  before_action :set_assignment, only: [:show, :edit, :update, :destroy,
+                                        :last_submitted_hours_and_feedbacks]
   before_action :set_volunteer, only: [:find_client]
```
casaper: I would say it is more explicit to call `:set_volunteer` within `find_client`, since you're only calling it for that action ... same for the `before_action :set_client, only: [:find_volunteer]`

app/controllers/hours_controller.rb
```diff
@@ -1,5 +1,5 @@
 class HoursController < ApplicationController
-  before_action :set_hour, only: [:show, :edit, :update, :destroy]
+  before_action :set_hour, only: [:show, :edit, :update, :destroy, :create_redirect]
```
casaper: the before action should  rather run at `:create` instead of `:create_redirect`, because thats the context where you want hour to be set as class variable

app/controllers/feedbacks_controller.rb
casaper: or maybe another option to get `GroupOffer` type paths:

```ruby
GroupAssignment.find_by(group_offer: @feedbackable, volunteer: @volunteer)
```

casaper: Oh, actually you'd need to use where and .last, because in case there is more than one `GroupAssignment` matching, you'd want to grab the later.

app/models/assignment.rb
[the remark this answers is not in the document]
casaper: Actually that was just bullshit... Sequel will not know what feedbackable is... 🤦‍♂️ 

but anyhow:

Solving what is possible within a scope inside `Feedback` would be the way to go.
In this case it could also be somewhat in that manner:

```ruby
scope :some_name, lambda { |submitted_at, volunteer_user|
  if submitted_at.present?
    where('created_at > ?', feedbackable.submitted_at).where(author: volunteer_user)
  else
    where(author: volunteer_user)
  end
}
```

app/controllers/hours_controller.rb
```diff
@@ -49,6 +50,20 @@ def set_volunteer
...
+    if session[:request_url].include?('last_submitted_hours_and_feedbacks')
+      if hourable.class == Assignment
+        redirect_to last_submitted_hours_and_feedbacks_assignment_path(hourable), make_notice
+      else
+        group_assignment = hourable.group_assignments.where(volunteer: @volunteer).last
+        redirect_to last_submitted_hours_and_feedbacks_group_assignment_path(group_assignment), make_notice
+      end
+    else
+      redirect_to @volunteer, make_notice
+    end
```
toupeira: You could remove the nested conditions to make it a bit more readable:

```ruby
if !session[:request_url].include?('last_submitted_hours_and_feedbacks')
  redirect_to @volunteer, make_notice
elsif hourable.class == Assignment
  ...
else
  ...
end
```

app/models/assignment.rb
```diff
@@ -70,4 +70,20 @@ def to_label
     label += " - #{period_start && I18n.l(period_start)}"
     label
   end
+
+  def hours_since_last_submitted
+    if submitted_at.present?
```
toupeira: `submitted_at?`

app/views/application/_last_submitted_hours_and_feedbacks.html.slim
```diff
@@ -0,0 +1,59 @@
...
+    - feedbacks.each do |record|
+      tr
+        td= record.feedbackable.to_label
+        td= record.goals
+        td= record.achievements
+        td= record.future
+        td= record.comments
+        td= t(record.conversation)
+
+- if confirmable.class == Assignment
```
toupeira: Not sure what class `confirmable` is, can you maybe use `.assignment?`

app/views/assignments/_form.html.slim
```diff
@@ -1,34 +1,26 @@
...
-        = f.input :state, collection: Assignment::STATES, default: :suggested
-
   .row
     .col-xs-12
-      = f.input :confirmation
+      - if action_name == 'new' && !params[:client_id]
+        = f.association :client, include_blank: true, collection: Client.need_accompanying
+      - else
+        = f.association :client, include_blank: true,
+          selected: params[:client_id] || (@assignment.client.id if @assignment.client.present?)
```
toupeira: `params[:client_id] || @assignment&.client.id`

app/views/assignments/_form.html.slim
```diff
@@ -1,34 +1,26 @@
...
+  .row
+    .col-xs-12= f.input :period, as: :date_range_picker, label: false
   .row
     .col-xs-12
-      = f.button :submit
+      - if action_name == 'new' && !params[:volunteer_id]
+        = f.association :volunteer, include_blank: true, collection: Volunteer.seeking_clients
+      - else
+        = f.association :volunteer, include_blank: true,
+          selected: params[:volunteer_id] || (@assignment.volunteer.id if @assignment.volunteer.present?)
```
toupeira: `params[:client_id] || @assignment&.volunteer.id`

app/controllers/hours_controller.rb
chrysalag: Not so sure, because before `:create` hour is not saved (so find crashes). When `:create_redirect` is called, hour is already saved.

app/controllers/feedbacks_controller.rb
chrysalag: Maybe I am a bit confused, because I don't see why that is simpler.

app/views/assignments/_form.html.slim
chrysalag: Combined with the next suggested change that results in validation error, `Client must exist, Volunteer must exist.`

app/controllers/hours_controller.rb
casaper: true

app/views/assignments/_form.html.slim
casaper: what is with this?
```ruby
f.association ... selected: @assignment&.client.id || params[:client_id]
```

would that work?

app/controllers/assignments_controller.rb
```diff
@@ -59,19 +58,35 @@ def destroy
...
   end
 
+  def last_submitted_hours_and_feedbacks
+    @last_submitted_hours = @assignment.hours_since_last_submitted
+    @last_submitted_feedbacks = @assignment.feedbacks_since_last_submitted
+  end
+
   private
 
+  def update_redirect
```
casaper: Actually, to be honest, that whole thing with the redirect is kind of difficult. 


Why not create an own action, because it wouldn't need anything unsafe with referers, plus you'd not need to pass get variables with that form button at all. In fact you could just make it a link with the method put, like the delete links we had. So no form needed.

```ruby
# AssignmentsController
def mark_viewed
   # Do your mark viewed stuff and have full control, not accepting some get parmeters maybe not in your control
   redirect_to back_to_your_feedback_list, notice: what ever special notice ou want to use
end

# config/routes.rb

resources :volunteers do
    # ...
    resources :group_offers do
      resources :feedbacks do
        put :mark_viewed, on: :member
      end
      resources :trial_feedbacks do
        put :mark_viewed, on: :member
      end
    end
    resources :assignments do
      resources :feedbacks do do
        put :mark_viewed, on: :member
      end
      resources :trial_feedbacks do
        put :mark_viewed, on: :member
      end
    end
  end
```

or even make it shorter with a routes concern:

```ruby
# routes.rb
concern :mark_view do
   put :mark_viewed, on: :member
end

resources :volunteers do
    # ...
    resources :group_offers do
      resources :feedbacks, concerns: :mark_view
      resources :trial_feedbacks, concerns: :mark_view
    end
    resources :assignments do
      resources :feedbacks, concerns: :mark_view
      resources :trial_feedbacks, concerns: :mark_view
    end
  end
```

app/controllers/feedbacks_controller.rb
```diff
@@ -24,7 +25,7 @@ def create
     @feedback.feedbackable = @feedbackable
     authorize @feedback
     if @feedback.save
-      redirect_to @feedback.volunteer, make_notice
+      redirect_to create_redirect, make_notice
```
casaper: And for this you could simply pass a get param to the 
`/volunteers/99/assignments/99/feedbacks/new?last_submitted=true`
in the `/app/views/feedbacks/_form.html.slim` you could then add this get parameter to the action if existing, and then your controller `create` action will know where its comming from - where to redirect to. Or at least without relying on refferers.

app/controllers/hours_controller.rb
```diff
@@ -49,6 +50,18 @@ def set_volunteer
     @volunteer = Volunteer.find(params[:volunteer_id]) if params[:volunteer_id]
   end
 
+  def create_redirect
+    hourable = @hour.hourable
```
casaper: What is the advantage of this to that:
```ruby
    if !session[:request_url].include?('last_submitted_hours_and_feedbacks')
      @volunteer
    elsif @hour.assignment?
      last_submitted_hours_and_feedbacks_assignment_path(@hour.hourable)
    else
      group_assignment = @hour.hourable.group_assignments.where(volunteer: @volunteer).last
      last_submitted_hours_and_feedbacks_group_assignment_path(group_assignment)
    end
```

app/views/assignments/_form.html.slim
chrysalag: Nope. `undefined method `id' for nil:NilClass`

app/views/application/_last_submitted_hours_and_feedbacks.html.slim
chrysalag: `confirmable` is either an `Assignment` or a `GroupAssignment`, that is defined in the view that calls the template. `assignment?` method is not available in `Assignment` or `GroupAssignment`

app/views/assignments/_form.html.slim
casaper: `params[:volunteer_id] || @assignment&.volunteer&.id`

that works. I tryed it. `@assignment&.volunteer.id` throws `NoMethodError: undefined method `id' for nil:NilClass` but `@assignment&.volunteer&.id` doesn't

PR #382, opened 2017-12-08
Changes: client index and form

chrysalag:
# [Story in Trello](https://trello.com/c/TdC90yra/91-klienten-view-%C3%A4nderungen)

### Done for approval?

* ~[ ] Story under test~
* ~[ ] Mobile works~
* ~[ ] Seeds created~
* [x] Code is reviewed

### Done after the merge?


test/system/assignments_test.rb
```diff
@@ -28,7 +28,10 @@ class AssignmentsTest < ApplicationSystemTestCase
   test 'assign unassigned client - client side' do
```
nuritnt: This whole test is not needed anymore, since there shouldn't be any matching going from the client on.
Which also means the buttons `Freiwillige finden` and `show Assignment` shouldn't be needed anymore. But to be sure I will have more clarification on that today after the meeting with her.

PR #385, opened 2017-12-13
Changes: Assignments

chrysalag:
# [Story in Trello](https://trello.com/c/0CZeXxo3/102-begleitungen-%C3%A4nderungen)

### Done for approval?

* [x] Story under test
* ~[ ] Mobile works~
* ~[ ] Seeds created~
* [x] Code is reviewed

### Done after the merge?

### DIFF
--- a/app/views/assignments/_assignment.html.slim
+++ b/app/views/assignments/_assignment.html.slim
@@ -1,11 +1,10 @@
 tr
+  td = link_to assignment.volunteer, assignment.volunteer if action_index?
   td
     - if policy(Client).show?
       = link_to assignment.client, assignment.client
     - else
       = assignment.client
-  - if action_name == 'index'
-    td = link_to assignment.volunteer, assignment.volunteer
   td= t("simple_form.options.assignment.state.#{assignment.state}") if assignment.state.present?
   td= l(assignment.period_start) if assignment.period_start
   td= l(assignment.period_end) if assignment.period_end
--- a/app/views/assignments/_assignment.html.slim
+++ b/app/views/assignments/_assignment.html.slim
@@ -7,7 +7,8 @@ tr
   - if action_name == 'index'
     td = link_to assignment.volunteer, assignment.volunteer
   td= t("simple_form.options.assignment.state.#{assignment.state}") if assignment.state.present?
-  td= l(assignment.created_at, format: :date_only)
+  td= l(assignment.period_start) if assignment.period_start
+  td= l(assignment.period_end) if assignment.period_end
   td
     - if policy(User).show?
       = link_to assignment.creator.navigation_name, profile_path(assignment.creator) if assignment.creator.present?
--- a/app/views/assignments/_assignment.html.slim
+++ b/app/views/assignments/_assignment.html.slim
@@ -11,9 +11,9 @@ tr
   td= l(assignment.period_end) if assignment.period_end
   td
     - if policy(User).show?
-      = link_to assignment.creator.navigation_name, profile_path(assignment.creator) if assignment.creator.present?
+      = link_to assignment.creator.full_name, profile_path(assignment.creator) if assignment.creator.present?
     - else
-      = assignment.creator.navigation_name
+      = assignment.creator.full_name
   = render 'index_actions', subject: assignment
   - if current_user.volunteer == assignment.volunteer
     td.hidden-print= link_to t_title(:edit, Assignment), edit_assignment_path(assignment)
--- a/app/views/assignments/_assignment.html.slim
+++ b/app/views/assignments/_assignment.html.slim
@@ -10,8 +10,8 @@ tr
   td= l(assignment.period_start) if assignment.period_start
   td= l(assignment.period_end) if assignment.period_end
   td
-    - if policy(User).show?
-      = link_to assignment.creator.full_name, profile_path(assignment.creator) if assignment.creator.present?
+    - if policy(User).show? && assignment.creator.profile.present?
+      = link_to assignment.creator.full_name, profile_path(assignment.creator)
     - else
       = assignment.creator.full_name
   = render 'index_actions', subject: assignment
--- a/app/views/assignments/_assignment.html.slim
+++ b/app/views/assignments/_assignment.html.slim
@@ -10,9 +10,9 @@ tr
   td= l(assignment.created_at, format: :date_only)
   td
     - if policy(User).show?
-      = link_to assignment.creator, assignment.creator if assignment.creator.present?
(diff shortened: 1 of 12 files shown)


app/views/assignments/_assignment.html.slim
```diff
@@ -7,12 +7,13 @@ tr
   - if action_name == 'index'
     td = link_to assignment.volunteer, assignment.volunteer
   td= t("simple_form.options.assignment.state.#{assignment.state}") if assignment.state.present?
-  td= l(assignment.created_at, format: :date_only)
+  td= l(assignment.period_start) if assignment.period_start
+  td= l(assignment.period_end) if assignment.period_end
   td
     - if policy(User).show?
-      = link_to assignment.creator, assignment.creator if assignment.creator.present?
+      = link_to assignment.creator.full_name, profile_path(assignment.creator) if assignment.creator.present?
```
nuritnt: why is a check here needed if it exists? If a creator would be existing, he couldn't create even an assignment, or did I get this logic wrong?^^

config/locales/de.yml
```diff
@@ -254,6 +254,7 @@ de:
         <<: *id-generic_keys
         amount: Betrag
         assignment: Begleitung
+        created_by: Verantwortliche/r
```
nuritnt: where is this still used? In general for created_by should be used "angemeldet von"

chrysalag: At billing expenses

nuritnt: ah true!

app/views/assignments/_assignment.html.slim
chrysalag: Maybe it is in case the user gets deleted

app/views/assignments/_assignment.html.slim
```diff
@@ -7,12 +7,13 @@ tr
   - if action_name == 'index'
     td = link_to assignment.volunteer, assignment.volunteer
   td= t("simple_form.options.assignment.state.#{assignment.state}") if assignment.state.present?
-  td= l(assignment.created_at, format: :date_only)
+  td= l(assignment.period_start) if assignment.period_start
+  td= l(assignment.period_end) if assignment.period_end
   td
     - if policy(User).show?
-      = link_to assignment.creator, assignment.creator if assignment.creator.present?
+      = link_to assignment.creator.full_name, profile_path(assignment.creator.profile) if assignment.creator.present?
```
casaper: `if assignment.creator.present?` can actually be dropped, as an Assignment would never be valid without having one.
It could be soft deleted tho, but that case is caught by:

```ruby
belongs_to :creator, -> { with_deleted }, class_name: 'User'
```

So as long as our db is not corrupt, we can always, even if softdeleted, expect Assignment to have a creator.

app/views/assignments/_assignment.html.slim
```diff
@@ -7,12 +7,13 @@ tr
...
     - else
-      = assignment.creator
+      = assignment.creator.full_name
```
casaper: `assignment.creator.full_name` on the other hand could fail on edge case, because `User` could have its Profile record missing.
I'd suggest `assignment.creator&.full_name` or even `assignment.creator&.profile&.contact&.full_name` because I wouldn't be 100% sure how this delegate on `User` acts in that case.

app/views/volunteers/_columns.html.slim
```diff
@@ -20,7 +20,7 @@
         th= sort_link @q, :interests
         th= sort_link @q, :acceptance
         th= sort_link @q, :external
-        th= t('registered_by')
+        th= t_attr(:created_by)
         th colspan=(action_name == 'index' ? '4' : '2')
```
casaper: we have `action_index?` now.

`app/helpers/action_name_helper.rb`

app/views/assignments/_assignment.html.slim
```diff
@@ -7,12 +7,13 @@ tr
...
-  td= l(assignment.created_at, format: :date_only)
+  td= l(assignment.period_start) if assignment.period_start
+  td= l(assignment.period_end) if assignment.period_end
   td
     - if policy(User).show?
-      = link_to assignment.creator, assignment.creator if assignment.creator.present?
+      = link_to assignment.creator.full_name, profile_path(assignment.creator) if assignment.creator.present?
     - else
-      = assignment.creator
+      = assignment.creator.full_name
```
casaper: assignment.creator.full_name on the other hand could fail on edge case, because User could have its Profile record missing.
I'd suggest assignment.creator&.full_name or even assignment.creator&.profile&.contact&.full_name because I wouldn't be 100% sure how this delegate on User acts in that case.

app/views/assignments/_assignment.html.slim
casaper: `if assignment.creator.present?` can actually be dropped, as an `Assignment` would never be valid without having one.
It could be soft deleted tho, but that case is caught by:
```ruby
belongs_to :creator, -> { with_deleted }, class_name: 'User'
```
So as long as our db is not corrupt, we can always (even if soft deleted) expect Assignment to have a creator.

app/views/assignments/_assignment.html.slim
chrysalag: Based on how `full_name` is defined, wouldn't it load the email if the user has no profile?

`def full_name`
`  if profile&.contact`
`    "#{profile.contact.last_name}, #{profile.contact.first_name}"`
`  else`
`    email`
`  end`
`end`

app/views/assignments/_assignment.html.slim
chrysalag: I checked in the console and if the user gets deleted, the data for the creator of the assignment are kept and the assignment is valid. So this exact check is not needed, as even if it is deleted it will return true. What could go wrong is that if the user gets deleted the link won't work.

app/views/assignments/_assignment.html.slim
```diff
@@ -1,18 +1,18 @@
...
-  - if action_name == 'index'
-    td = link_to assignment.volunteer, assignment.volunteer
   td= t("simple_form.options.assignment.state.#{assignment.state}") if assignment.state.present?
-  td= l(assignment.created_at, format: :date_only)
+  td= l(assignment.period_start) if assignment.period_start
+  td= l(assignment.period_end) if assignment.period_end
   td
-    - if policy(User).show?
-      = link_to assignment.creator, assignment.creator if assignment.creator.present?
+    - if policy(User).show? && assignment.creator.profile.present?
```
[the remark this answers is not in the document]
chrysalag: If the user gets deleted, the link won't work.

